PR #1, opened 2017-12-13
Add Identity decorator prototype

markstory:
It will quack like the authentication one, but won't inherit from it because we don't want to add cross library dependencies. I tried to think of how we could support an interface that extends from cakephp/authentication and couldn't think of an approach that didn't require multiple inheritance (which we don't have).

This is just a sketch to get discussions started and try and move forwards more with how the decorator should work.


src/AuthorizationServiceInterface.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * Check whether the provided user can perform an action on a resource.
+     *
+     * @param \Authorization\IdentityInterface $user The user to check permissions for.
+     * @param string $action The action/operation being performed.
+     * @param mixed $resource The resource being operated on.
+     * @return bool
+     */
+    public function can(IdentityInterface $user, $action, $resource);
```
markstory: Not sure if the service should typehint here.

robertpustulka: I'd relax this parameter and allow anything to be passed.

src/IdentityDecorator.php
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Delegate unknown methods to decorated identity.
+     *
+     * @param string $method The method being invoked.
+     * @param arry $args The arguments for the method.
+     * @return mixed
+     */
+    public function __call($method, $args)
+    {
+        $call = [$this->identity, $method];
```
robertpustulka: Needs a `is_object` check probably, as the identity may be an array.

src/IdentityDecorator.php
```diff
@@ -0,0 +1,135 @@
...
+        $this->authorization = $service;
+        $this->identity = $identity;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function can($action, $resource)
+    {
+        return $this->authorization->can($this, $action, $resource);
```
robertpustulka: Should `$this` be passed to the service, or maybe `$this->identity`?
Policies may typehint some internal identity object or do `instance_of` checks.

src/AuthorizationServiceInterface.php
markstory: Done!

src/IdentityDecorator.php
markstory: That is a good point. I was concerned that passing the wrapped identity would make combining permissions/policy checks defined in other resources harder. For example passing the unwrapped user would make something like the following not possible.

```php
function update($user, $thing)
{
  return $user->can('update', $this->category);
}
```

src/IdentityDecorator.php
markstory: done.

src/IdentityDecorator.php
robertpustulka: We can have some kind of `getOriginal` method in the identity. This way we could combine checks in policies and still access the original object for some internal checks.

For example:
```php
function canUpdate($identity, $object) {
   return $identity->can(‘access’, $object) || $identity->getOriginal() instanceof Admin);
}

markstory: I can add a getOriginal() method to the interface and decorator.

PR #3, opened 2017-12-15
Policy Resolver prototype

robertpustulka:
Added an interface for policy resolvers and a basic map resolver.

### DIFF
--- a/src/Policy/BeforePolicyInterface.php
+++ b/src/Policy/BeforePolicyInterface.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authorization\Policy;
+
+use Authorization\IdentityInterface;
+
+interface BeforePolicyInterface
+{
+
+    /**
+     * Defines a pre-authorization check.
+     *
+     * @param \Authorization\IdentityInterface $identity Identity object.
+     * @param mixed $resource The resource being operated on.
+     * @return bool
+     */
+    public function before(IdentityInterface $identity, $resource);
+}
--- a/src/Policy/BeforePolicyInterface.php
+++ b/src/Policy/BeforePolicyInterface.php
@@ -16,6 +16,10 @@ namespace Authorization\Policy;
 
 use Authorization\IdentityInterface;
 
+/**
+ * This interface should be implemented if a policy class needs to perform a
+ * pre-authorization check before the action access check takes place.
+ */
 interface BeforePolicyInterface
 {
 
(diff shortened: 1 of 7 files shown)


src/Policy/BeforePolicyInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authorization\Policy;
+
+use Authorization\IdentityInterface;
+
+interface BeforePolicyInterface
```
markstory: We should have a doc block on when and why you might want to use this interface.

src/Policy/MapResolver.php
```diff
@@ -0,0 +1,89 @@
...
+     * @throws \InvalidArgumentException When a policy class does not exist.
+     */
+    public function map($resourceClass, $policyClass)
+    {
+        if (!class_exists($policyClass)) {
+            $message = sprintf('Policy class `%s` does not exist.', $policyClass);
+            throw new InvalidArgumentException($message);
+        }
+
+        $this->map[$resourceClass] = $policyClass;
```
markstory: Do we need to make sure the resourceClass exists as well?

src/Policy/MapResolver.php
```diff
@@ -0,0 +1,89 @@
...
+            throw new InvalidArgumentException($message);
+        }
+
+        $class = get_class($resource);
+
+        if (!isset($this->map[$class])) {
+            throw new MissingPolicyException([$class]);
+        }
+
+        return new $this->map[$class];
```
markstory: Should we support closures for creating policy instances?

src/Policy/MapResolver.php
robertpustulka: I’m still not sure if the plugin should grant access to non-object resources. Does it even make sense?

src/Policy/MapResolver.php
robertpustulka: Policy object should define `canSomething` methods. Closures can provide that. Or do you have something different in mind?

markstory: I was thinking something like

```php
$resolver-map([Thing::class => function($resolver) {
  return new ThingPolicy(...);
}]);
```

robertpustulka: Good idea :+1:

An instantiated policy object could be passed as well

src/Policy/ResolverInterface.php
```diff
@@ -0,0 +1,27 @@
...
+interface ResolverInterface
+{
+
+    /**
+     * Resolves the policy object based on the authorization resource.
+     *
+     * @param mixed $resource A resource that the access is checked against.
+     * @return object
+     */
+    public function getPolicy($resource);
```
robertpustulka: Should we have an `$action` as a second parameter? Somebody may want to map closures or invokable objects as policies for individual actions.

This would require having a policy handler. Default handler would call `canAction()` method on a policy, but somebody might implement a hadler that would invoke the policy object or call other method. The similar design is implemented by tactician: https://tactician.thephpleague.com/tweaking-tactician/

markstory: > Should we have an $action as a second parameter? Somebody may want to map closures or invokable objects as policies for individual actions.

I don't think that is necessary. We've been planning on mapping policies to resources. Including actions muddies the purpose of the action parameter when running a policy.

src/Policy/MapResolver.php
```diff
@@ -84,6 +102,16 @@ public function getPolicy($resource)
...
+
+        if (is_callable($policy)) {
+            return $policy($resource, $this);
+        }
+
+        if (is_object($policy)) {
+            return $policy;
+        }
+
+        return new $policy();
```
markstory: Nice :+1:

PR #5, opened 2017-12-18
Add ORM Resolver

markstory:
Add a resolver that finds policy classes for ORM objects based on conventions. I'll add documentation for both resolver classes in a follow up pull request.


composer.json
```diff
@@ -20,24 +20,21 @@
...
 	},
 	"authors": [
-		{
-			"name": "Florian Krämer",
-			"homepage": "http://florian-kraemer.net",
-			"role": "Developer"
-		}
+			{
+					"name": "CakePHP Community",
+					"homepage": "https://github.com/cakephp/cakephp/graphs/contributors"
```
robertpustulka: Should it point to the plugin's contributors?

src/Policy/OrmResolver.php
```diff
@@ -0,0 +1,129 @@
...
+     * @throws \InvalidArgumentException When a resource is not an ORM object.
+     * @throws \Authorization\Policy\Exception\MissingPolicyException When a policy for the
+     *   resource has not been defined.
+     */
+    public function getPolicy($resource)
+    {
+        if ($resource instanceof EntityInterface) {
+            return $this->getEntityPolicy($resource);
+        }
+        if ($resource instanceof RepositoryInterface) {
```
robertpustulka: Should we support `QueryInterface` as well? This would be helpful for decorating a query object.  The policy could be resolved based on the query repository.

tests/TestCase/Policy/OrmResolverTest.php
```diff
@@ -0,0 +1,94 @@
...
+use TestApp\Policy\TestPlugin\BookmarkPolicy;
+use TestPlugin\Model\Entity\Bookmark;
+use TestPlugin\Model\Entity\Tag;
+use TestPlugin\Policy\TagPolicy;
+
+class OrmResolverTest extends TestCase
+{
+    public function testGetPolicyUnknownObject()
+    {
+        $this->setExpectedException(InvalidArgumentException::class);
```
robertpustulka: This method is deprecated and missing in PHPUnit 6.0, use `expectException()` instead.

composer.json
markstory: Yes. This is copy paste fail.

src/Policy/OrmResolver.php
markstory: I was considering it. Would you expect the policy to resolve to the table policy, or something else?

tests/TestCase/Policy/OrmResolverTest.php
markstory: Fixed now.

src/Policy/OrmResolver.php
robertpustulka: Maybe this could resolve to `ArticlesQuery` policy and fall back to `ArticlesTable`. Would that make sense or bring too much confusion?

markstory: We could have `ArticlesQueryPolicy`. Relying on policies to mutate queries in place feels kind of gross though. In pundit 'query policies' are a special type of policy that accept a query and return a new query with conditions applied. An example would be.

```ruby
query = policy_scope(Bookmarks.find).resolve
```

The `resolve` method update the query with additional conditions based on the current user.

robertpustulka: Maybe we should have another method in the service for resource modification based on policy rules? Running a query through `can()` method in order to inject some conditions doesn’t seem intuitive.

markstory: I agree that `can()` mutating state is icky feeling. I can draft up some ideas and post them in an issue later this week.

markstory:
@robertpustulka Does this look if Query policies are handled separately?

robertpustulka:
@markstory I was thinking about something like:
`$query = $user->authorize($articles->find(), 'index')`

Which would call `ArticlesQueryPolicy::authorizeIndex()`

PR #6, opened 2017-12-18
Add AuthorizationService

robertpustulka:
I restored the `IdentityInterface` typehint to `AuthorizationServiceInterface::can()`. The `IdentityInterface` exposes the original data now.


src/AuthorizationService.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Returns a policy action handler.
+     *
+     * @param object $policy Policy object.
+     * @param string $action Action name.
+     * @return callable
+     * @throws \Authorization\Policy\Exception\MissingMethodException
+     */
+    protected function getHandler($policy, $action)
```
robertpustulka: Should this be abstracted? I have something like a `PolicyHandlerInterface` in mind. Implementations could call methods on policy objects in many way. This could be helpful if someone was porting their policies from another backend.

markstory: I think until we have a use-case we shouldn't add extension points as they may not turn out to be that useful.

With that said, this protected method makes sense and keeps `can()` short and easy to read.

tests/TestCase/AuthorizationServiceTest.php
```diff
@@ -0,0 +1,135 @@
...
+    public function testCan()
+    {
+        $resolver = $this->createMock(ResolverInterface::class);
+        $entity = new Article();
+        $policy = new ArticlePolicy();
+
+        $resolver->expects($this->once())
+            ->method('getPolicy')
+            ->with($entity)
+            ->willReturn($policy);
```
markstory: Instead of a mock you could use the `MapResolver`. It would save LOC and give us integration tests for 'free'.

robertpustulka: I have a habit of testing thing in isolation as much as possibke, but it is a good point. I'll change this.

markstory:
Looks good. There is a conflict now though.

src/AuthorizationService.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function can(IdentityInterface $user, $action, $resource)
+    {
+        $policy = $this->resolver->getPolicy($resource);
+
+        if ($policy instanceof BeforePolicyInterface && !$policy->before($user, $resource)) {
```
markstory: Should we allow a policy's before() method to return true to allow all operations?

markstory: Should the before hook know the action being take as well?

robertpustulka: Maybe a before method could allow the authorization check if a policy method hasn't been defined.
If a method exists the result of that would be returned.

robertpustulka:
Conficts resolved.

robertpustulka:
I added the ability to return true if a before policy check passed but a method for action hasn't been defined.

This way for example the CRUD actions can fallback to the default check.

```php
class ArticlePolicy implements BeforePolicyInterface
{
    public function before($user, $article)
    {
        return $user->is_admin;
    }
}

$user->can('delete', $article); //returns true for admins, false for everyone else
```

src/AuthorizationService.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function can(IdentityInterface $user, $action, $resource)
+    {
+        $policy = $this->resolver->getPolicy($resource);
+
+        if ($policy instanceof BeforePolicyInterface) {
+            if (!$policy->before($user, $resource)) {
```
markstory: Should we only fail early if the `before` method returns false? That would allow

* `true` to skip the can method.
* `null` to fall through.
* `false` to abort early

 The use case I can think of for `true` is an administrator account which is allowed to do everything in the system.

robertpustulka:
I added the `$action` parameter to the before check.

PR #7, opened 2017-12-18
Added ResolverCollection

robertpustulka:
Useful for aggregating multiple resolvers.

### DIFF
--- a/src/Policy/ResolverCollection.php
+++ b/src/Policy/ResolverCollection.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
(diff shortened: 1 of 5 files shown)


markstory:
I was thinking of adding this exact same thing :+1:

src/Policy/ResolverCollection.php
```diff
@@ -0,0 +1,68 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authorization\Policy;
+
+use Authorization\Policy\Exception\MissingPolicyException;
+
+class ResolverCollection implements ResolverInterface
```
markstory: Should include a short description of why/when one would use this class.

robertpustulka: Done

PR #14, opened 2017-12-21
Add a bake task for creating policies

markstory:
If the user has bake and authorization installed we can provide them with a task to generate policy classes. I've included stub methods for entities as I think they will be the most common type of policy created.

### DIFF
--- a/src/Template/Bake/Element/entity_methods.twig
+++ b/src/Template/Bake/Element/entity_methods.twig
@@ -1,0 +1,43 @@
+    /**
+     * Check if $user can create {{ name }}
+     *
+     * @param Authorization\IdentityInterface $user The user.
+     * @param {{ namespace }}\Model\Entity\{{ name }} ${{ variable_name }}
+     * @return bool
+     */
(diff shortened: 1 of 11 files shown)


src/Template/Bake/Element/entity_methods.twig
```diff
@@ -0,0 +1,43 @@
+    /**
+     * Check if $user can create {{ name }}
+     *
+     * @param Authorization\IdentityInterface $user The user.
+     * @param {{ namespace }}\Model\Entity\{{ name }} ${{ variable_name }}
+     * @return bool
+     */
+    public function canCreate(IdentityInterface $user, ${{ variable_name }})
```
robertpustulka: Should the entity be type hinted as entity class name or `EntityInterface`?

markstory: I can add typehints for the entity classes.

markstory:
@robertpustulka @burzum Is this good to go?

robertpustulka:
Looks OK

PR #15, opened 2017-12-25
Draft Authorization component

markstory:
Adding a draft of the authorization component that only adds a single method `authorize()` which works like `can()` but raises exceptions allowing controller actions to be terse and not contain error handling.

I've not exposed can() directly as I think authorize() is a better shortcut to use at this layer. I've also fixed a mistake in the OrmResolver where the wrong kind of exception was being raised.


burzum:
It looks OK for me.

markstory:
I will get started on query scopes next.

PR #17, opened 2017-12-31
Implement authorization scopes.

markstory:
Allow authorization policies to mutate objects in place. This enables query objects to have conditions applied or pre-compute authorization checks on a collection of objects.


robertpustulka:
Nice!

We probably need a resolver for Query objects as discussed in another PR.

markstory:
> We probably need a resolver for Query objects as discussed in another PR.

I was planning on adding that next.

PR #19, opened 2018-01-05
Support model authorization in controllers.

robertpustulka:


tests/TestCase/Controller/Component/AuthorizationComponentTest.php
```diff
@@ -104,4 +103,50 @@ public function testAuthorizeCustomException()
...
+
+    public function testAuthorizeModelSuccess()
+    {
+        $service = new AuthorizationService(new OrmResolver());
+        $identity = new IdentityDecorator($service, ['can_edit' => true]);
+        $this->Controller->request = $this->Controller->request
+            ->withAttribute('identity', $identity);
+
+        $this->Auth->authorizeModel();
+        $this->assertTrue(true);
```
markstory: This doesn't look right.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -86,4 +88,30 @@ protected function getIdentity(ServerRequestInterface $request)
...
     }
+
+    /**
+     * Model authorization handler.
+     *
+     * @return void
+     */
+    public function authorizeModel()
+    {
+        $this->authorize($this->getController()->loadModel());
```
markstory: For a view/edit/delete operation does is it useful to check if the user has access in general? Personally I would like to know if the current user is allowed to modify/access the specific records they are working on.

tests/TestCase/Controller/Component/AuthorizationComponentTest.php
robertpustulka: I’m not sure how to assert that exception has not been thrown.

src/Controller/Component/AuthorizationComponent.php
robertpustulka: The access for individual entities should be checked using entity policies IMO.

So `$this->Authorization->authorize($entity)` in action body. I’m not a fan of loading entities magically and checking access. Feels like a job for a plugin, Crud for instance.

robertpustulka: I can add a config option for supported actions if that’s your concern.

tests/TestCase/Controller/Component/AuthorizationComponentTest.php
markstory: You could assert the return of `authorizeModel`

src/Controller/Component/AuthorizationComponent.php
markstory: Yeah I don't want to magically load entities either. I could see this being useful for add at least.

tests/TestCase/Controller/Component/AuthorizationComponentTest.php
jeremyharris: What about `$this->assertNull($this->Auth->authorizeModel());`? Adds to assertion count, allows further assertions, and will fail if an exception is thrown.

robertpustulka:
I've added `actionMap` config option. It allows to map controller action names to authorization actions as well as disabling authorization checks on certain actions. By default the controller action name is used.

```php
$this->loadComponent('Authorization, ['actionMap' => [
    'edit' => 'modify' //will use `canModify` auth check
    'index' => false //will not check access for `index` action
]];
```

Issue #21, opened 2018-01-08
[RFC] Support plugin map in OrmResolver


robertpustulka:
I have a use case, where I have my app logic in a plugin (let's cal it a logic plugin) and use there some models from another plugin (call it data plugin).

Currently the only way to define policies for my models is to:
- add policies to the "data plugin" (not possible in my case)
- add policies to my app (breaks my design, where logic is defined in a logic plugin)
- use map resolver and map each "data plugin" model to the policy in a "logic plugin" (makes little sense)
- extend each "data plugin" model in "logic plugin" model and use those instead (still lots of senseless work)

My idea is to have the ability to map plugins in a resolver so the policies would be looked for in another plugin. For example:
```php
new OrmResolver([
    'appNamespace' => 'MyApp',
    'pluginMap' => [
        'DataPlugin' => 'LogicPlugin'
    ]
]);
```

What is your opinion on this?

markstory:
Seems reasonable to me.

robertpustulka:
Should the policies for DataPlugin be looked for in `LogicPlugin\Policy\DataPlugin` as the app overrides or just in the root namespace?

markstory:
I don't know. But it seems excessive to have overrides in an override.

PR #23, opened 2018-01-09
Add support for plugin name overrides

robertpustulka:
Implements #21


src/Policy/OrmResolver.php
```diff
@@ -34,20 +33,29 @@ class OrmResolver implements ResolverInterface
      */
     protected $appNamespace;
 
+    /**
+     * Plugin name overrides.
+     *
+     * @var array<string, string>
+     */
+    protected $overrides;
```
markstory: Should this be initialized to an empty array?

PR #27, opened 2018-01-14
Add ability to ensure authorization was checked.

markstory:
I've found this to be a helpful feature in other libraries during development/testing. It helps you ensure that your controller actions all check authorization. Failing to check authorization will trigger an exception which will fail the test.

If we add a way to 'skip' authorization in the future it should also modify this flag as skipping authorization is a type of 'check'.


robertpustulka:
Good idea :+1:

PR #29, opened 2018-01-17
Move `IdentityInterface` instance check to `buildIdentity()`.

robertpustulka:
Current implementation does not allow to call custom callable decorators
on existing `IdentityInterface` instances. This makes it impossible to
easily inject authorization service into custom identity instances.

### DIFF
--- a/tests/TestCase/Middleware/AuthorizationMiddlewareTest.php
+++ b/tests/TestCase/Middleware/AuthorizationMiddlewareTest.php
@@ -29,6 +29,7 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
 use stdClass;
+use TestApp\Identity;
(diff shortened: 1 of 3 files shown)


markstory:
Any test we could add?

robertpustulka:
Yes, I will add some tests.

tests/TestCase/Middleware/AuthorizationMiddlewareTest.php
```diff
@@ -255,6 +256,38 @@ public function testCustomIdentity()
...
+                $identity->setService($service);
+
+                return $identity;
+            },
+            'requireAuthorizationCheck' => false,
+        ]);
+
+        $identity->expects($this->once())
+            ->method('setService')
+            ->with($service);
```
markstory: Instead of mocking your stub class you could implement the actual method and rely on fewer mocks.

Issue #25, opened 2018-01-11
Restricting access for not logged in users


robertpustulka:
Do we wan't to support some replacement for the `AuthComponent:allow()`/`deny()` methods?

Currently `AuthorizeComponent` requires that the identity is present in a request. There's no build-in way of checking if an identity is present and take some action if it is not (ie redirect to login page or throw a forbidden exception).

Not sure if this feature should belong here or maybe Authentication plugin?

markstory:
How would you want that to be expressed in the controller action? Would something like 

```php
// Throw exception when no user is present
$this->Authorization->requireUserPresent()
```
work?

robertpustulka:
I think that the current core Auth API is OK. We could port that to the plugin.

```php
// UsersController.php
public function initialize()
{
    ...
    $this->Authorization->deny('*');
    $this->Authorization->allow('login');
}
```

robertpustulka:
Implemented by #30

PR #30, opened 2018-01-25
Add authorization bypassing

robertpustulka:
This is an implementation for #25.

By default all requests require authorization check when `requireAuthorizationCheck` is set to true.
This PR adds the ability to skip authorization by calling `skipAuthorization` on service or component without executing policy check.

`AuthorizationComponent` allows for automatic authorization skipping as well.
This could be achieved by marking actions in `skipAuthorization` config array, thus making actions publicly accessible or not.

For example:
```php
$this->loadComponent('Authorization.Authorization', [
    'skipAuthorization' => [
        'login'
    ]
];
```

I refactored model authorization config to use the same approach.

Mapping action has changed it responsibility a bit. Now all authorization checks made on component without providing action name explicitly will use mapped action names. This also means that setting action names to true/false will no longer work as enabling/disabling actions has been moved to `authorizeModel` config. This makes using the map and automatic authorization simplier in my mind.

I updated the docs as well, the changes are exaplained there.

After this gets accepted and merged, I'll implement handlers for unathorized requests so for instance all `MissingIdentityException`s could be redirected to the login page.

### DIFF
--- a/src/Controller/Component/AuthorizationComponent.php
+++ b/src/Controller/Component/AuthorizationComponent.php
@@ -14,7 +14,9 @@
  */
 namespace Authorization\Controller\Component;
 
+use Authorization\AuthorizationServiceInterface;
 use Authorization\Exception\ForbiddenException;
+use Authorization\Exception\MissingIdentityException;
 use Authorization\IdentityInterface;
 use Cake\Controller\Component;
 use Cake\Http\ServerRequest;
@@ -44,7 +46,11 @@ class AuthorizationComponent extends Component
      */
     protected $_defaultConfig = [
         'identityAttribute' => 'identity',
+        'serviceAttribute' => 'authorization',
         'authorizationEvent' => 'Controller.initialize',
+        'skipAuthorization' => [
+            self::ALL => false,
+        ],
         'authorizeModel' => [
             self::ALL => true,
         ],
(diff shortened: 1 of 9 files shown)


src/Controller/Component/AuthorizationComponent.php
```diff
@@ -31,16 +34,26 @@
 class AuthorizationComponent extends Component
 {
 
+    /**
+     * Constant for all actions config key.
+     */
+    const ALL = '*';
```
markstory: I worry that this may lead to people accidentally opening up their actions by accident. Forcing the use of a whitelist forces people to be specific on which actions do not have authorization checks applied.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -59,12 +72,11 @@ public function authorize($resource, $action = null)
...
         $request = $this->getController()->request;
         if ($action === null) {
-            $action = $request->getParam('action');
+            $action = $this->getDefaultAction($request);
         }
         $identity = $this->getIdentity($request);
         if (!$identity->can($action, $resource)) {
-            $class = $this->getConfig('forbiddenException');
-            throw new $class();
+            throw new ForbiddenException([$action, get_class($resource)]);
```
markstory: :+1: for this not being configurable.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -82,23 +94,67 @@ public function applyScope($resource, $action = null)
...
 
         return $identity->applyScope($action, $resource);
     }
 
+    /**
+     * Skips the authorization check.
+     *
+     * @return $this
+     */
+    public function skipAuthorization()
```
markstory: I like how simple this method is :+1:

src/Controller/Component/AuthorizationComponent.php
robertpustulka: It is a fair point. I'm wondering if we should do the same for model authorization. I think that model authorization should be done occasionally anyway (for methods that does not change state of existing entities or that query the db) so it would be easier to list only those actions and leave the rest for manual authorization.

markstory: I am :+1: for opt-in features for authorization. Keeping the potential surprise count low is important to me.

robertpustulka:
@markstory automatic authorization and skiping is now opt-in. Manual authorization is now required by default. I think this change will make the plugin less surprising to work with for devs :)

markstory:
Awesome stuff @robertpustulka 👏

PR #31, opened 2018-01-29
Unauthorized request handlers

robertpustulka:
Follow up to #25 and #30

This PR implements handlers for unauthorized requests.

Build in handlers (one for stringy URLs, the other for CakePHP Router) redirect requests (eg. to the login page) for a list of configured exception classes.

By default only the `MissingIdentityException` is redirected.


tests/TestCase/Middleware/Handler/RedirectHandlerTest.php
```diff
@@ -0,0 +1,93 @@
...
+            'queryParam' => null,
+        ]);
+
+        $this->assertEquals(302, $response->getStatusCode());
+        $this->assertEquals('/users/login', $response->getHeaderLine('Location'));
+    }
+
+    public function testHandleException()
+    {
+        $handler = new RedirectHandler;
```
dereuromark: We usually normalize with () for instantiation.

docs/Middleware-and-Component.md
```diff
@@ -60,6 +60,49 @@ $middlewareStack->add(new AuthorizationMiddleware($this, [
...
+* `exceptions` - a list of exception classes that should be redirected. By default only `MissingIdentityException` is redirected.
+* `queryParam` - the accessed request URL will be attached to the redirect URL query parameter (`url` by default).
+* `statusCode` - HTTP status code of a redirect, `302` by default.
+
+For example:
+
+```php
+$middlewareStack->add(new AuthorizationMiddleware($this, [
+    'unauthorizedHandler' => [
+        'className' => 'Authorization.Redirect',
```
dereuromark: Why not calling them what they are? StringUrlRedirect vs ArrayUrlRedirect maybe?
The CakeRedirect doesnt really say what it does (it only supports array, nothing else from what I read).

robertpustulka: Cake as a name suggests supports CakePHP :).

tests/TestCase/Middleware/Handler/RedirectHandlerTest.php
robertpustulka: Maybe we need a sniffer for those?

docs/Middleware-and-Component.md
```diff
@@ -60,6 +60,49 @@ $middlewareStack->add(new AuthorizationMiddleware($this, [
...
+
+* `Exception` - this handler will rethrow the exception, this is a default behavior of the middleware.
+* `Redirect` - this handler will redirect the request to the provided URL.
+* `CakeRedirect` - redirect handler with support for CakePHP Router.
+
+Both redirect handlers share the same configuration options:
+
+* `url` - URL to redirect to (`CakeRedirect` supports CakePHP Router syntax).
+* `exceptions` - a list of exception classes that should be redirected. By default only `MissingIdentityException` is redirected.
+* `queryParam` - the accessed request URL will be attached to the redirect URL query parameter (`url` by default).
```
markstory: Should we use `redirect` for compatibility with existing AuthComponent setups?

docs/Middleware-and-Component.md
```diff
@@ -60,6 +60,49 @@ $middlewareStack->add(new AuthorizationMiddleware($this, [
...
+        'queryParam' => 'redirectUrl',
+        'exceptions' => [
+            MissingIdentityException::class,
+            OtherException::class,
+        ],
+    ],
+]));
+```
+
+You can also add your own handler. Handlers should implement `Authorization\Middleware\Handler\HandlerInterface`,
```
markstory: Should we use a more specific name? Perhaps `FailureHandler`?

docs/Middleware-and-Component.md
```diff
@@ -60,6 +60,49 @@ $middlewareStack->add(new AuthorizationMiddleware($this, [
...
+be suffixed with `Handler` suffix and reside under your app's or plugin's 
+`Middleware\Handler` namespace.
+
+Configuration options are passed to the handler's `handle()` method as the last parameter.
+
+Handlers catch only those exceptions which extend `Authorization\Exception\Exception` class.
```
markstory: What about the whitelist? Could someone add `\RuntimeException` if they wanted?

src/Middleware/AuthorizationMiddleware.php
```diff
@@ -101,21 +104,44 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
 
-        $identity = $this->buildIdentity($service, $identity);
-
-        $request = $request->withAttribute($attribute, $identity);
-        $response = $next($request, $response);
+        try {
+            $response = $next($request, $response);
+        } catch (Exception $exception) {
+            $handler = $this->getHandler();
+            $response = $handler->handle($exception, $request, $response, (array)$this->getConfig('unauthorizedHandler'));
```
markstory: Should options be passed into the handler constructor? That would allow the objects to validate options earlier.

src/Middleware/Handler/HandlerFactory.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public static function create($name)
+    {
+        $class = App::className($name, 'Middleware/Handler', 'Handler');
+        if (!$class) {
+            $message = sprintf('Handler `%s` does not exist.', $name);
+            throw new RuntimeException($message);
+        }
+
+        $instance = new $class();
```
markstory: Pushing options into the constructor would let us use DI patterns more effectively.

docs/Middleware-and-Component.md
robertpustulka: Catching non-authorization exceptions seems out of the scope of this plugin. More like a general exception handler thing.

src/Middleware/Handler/HandlerFactory.php
robertpustulka: I wanted to avoid abstracting the constructor.

A similar approach has been implemented by authentication URL checkers.

docs/Middleware-and-Component.md
markstory: The code doesn't currently implement the check on `Authorization\Exception\Exception` though.

robertpustulka: Yes, this is intentional. I think that only missing identities should be redirected by default. Logged in users shouldn’t be redirected to the login page when they try to access unauthorized resource. This may cause a redirect loop. Also policy resolver exceptions shouldn’t be redirected as those are more like “internal error” exceptions.

src/Middleware/Handler/HandlerFactory.php
markstory: Ok. I guess method parameters are another form of DI.

docs/Middleware-and-Component.md
dereuromark: I don't get that - isn't everything here something for "Cake"?
After all:
```
"require": {
		"cakephp/core": "~3.5",
```

So clarying what those Handlers actually use seems to be more important than outlining that one is for Cake, and the other somewhat too :)

robertpustulka: > isn't everything here something for "Cake"?

@dereuromark No, it is a framework-agnostic plugin.

docs/Middleware-and-Component.md
markstory: Makes sense.

markstory:
Looks great :clap:

robertpustulka:
Do we need such high codecov diff hit target (99.66%) ?

markstory:
I think codecov generates the target based on previous builds.

PR #32, opened 2018-02-01
Fix unauthorized handler with authorization check.

robertpustulka:
Responses for unauthorized requests returned from handlers were not
returned by the middleware with authorization check enabled.

Authorization check has been moved to the `try` block so the authorization
check is made only when app does not throw any exception.

### DIFF
--- a/src/Middleware/AuthorizationMiddleware.php
+++ b/src/Middleware/AuthorizationMiddleware.php
@@ -111,12 +111,12 @@ class AuthorizationMiddleware
 
         try {
             $response = $next($request, $response);
+            if ($this->getConfig('requireAuthorizationCheck') && !$service->authorizationChecked()) {
+                throw new AuthorizationRequiredException(['url' => $request->getRequestTarget()]);
+            }
         } catch (Exception $exception) {
             $handler = $this->getHandler();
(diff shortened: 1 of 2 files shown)


src/Middleware/AuthorizationMiddleware.php
```diff
@@ -111,13 +111,13 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
 
         try {
             $response = $next($request, $response);
+            if ($this->getConfig('requireAuthorizationCheck') && !$service->authorizationChecked()) {
+                throw new AuthorizationRequiredException(['url' => $request->getRequestTarget()]);
```
markstory: Won't this result in a redirect when the redirect handler is used?

robertpustulka: Only when the handler is configured to catch this exception class. Redirect handler redirects only missing identities by default.

We could skip authentication in a catch block instead if accidental redirects are your concern.

markstory: Good point. I didn't consider that.

robertpustulka: This change will also allow to catch `AuthorizationRequiredException` if someone wants to handle it in any special way.

PR #34, opened 2018-03-13
Redirect targets should not include host names.

markstory:
Including host names sets us up for off domain redirects which is not good.

### DIFF
--- a/src/Middleware/UnauthorizedHandler/RedirectHandler.php
+++ b/src/Middleware/UnauthorizedHandler/RedirectHandler.php
@@ -93,7 +93,7 @@ class RedirectHandler implements HandlerInterface
     {
         $url = $options['url'];
         if ($options['queryParam'] !== null) {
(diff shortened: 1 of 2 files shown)


src/Middleware/UnauthorizedHandler/RedirectHandler.php
```diff
@@ -92,8 +92,8 @@ protected function checkException(Exception $exception, array $exceptions)
     protected function getUrl(ServerRequestInterface $request, array $options)
     {
         $url = $options['url'];
-        if ($options['queryParam'] !== null) {
-            $query = urlencode($options['queryParam']) . '=' . urlencode($request->getUri());
+        if ($options['queryParam'] !== null && $request->getMethod() === 'GET') {
```
markstory: I also found redirects on non-GET which is something to be avoided as those URLs could have side-effects or generate 405 errors.

PR #35, opened 2018-03-17
Docs updates & Property delegation

markstory:
Improve docs  & add __get delegation

* Show how to replace the decorator.
* Give better links.
* Supporting `__get` and `__isset` helps smooth out decorator usage as not being able to read properties is a pain when moving from a basic entity. It is also inconsistent with ArrayAccess interface.

### DIFF
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -31,6 +31,7 @@ use Authorization\Policy\OrmResolver;
 Then add the following to your `middleware()` method:
 
 ```php
+// Add authorization (after authentication if you are using that plugin too).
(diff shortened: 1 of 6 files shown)


docs/Quick-start-and-introduction.md
```diff
@@ -60,10 +61,22 @@ $this->loadComponent('Authorization.Authorization');
 ```
 
 By loading the [authorization component](./Component.php) we'll be able to check
-authorization on a per action basic more easily.
+authorization on a per action basic more easily. For example, we can do:
+
+```php
+public function edit($id)
```
dereuromark: Do we need `$id = null` here?

markstory: Sure.

PR #39, opened 2018-04-16
Adding RequestAuthorizationMiddleware

burzum:
This middleware is useful when you want to authorize your requests, for example each controller and action, against a role based access system or any other kind of authorization process that controls access to certain actions.

---

There are no tests yet on purpose, I would like to get feedback and finish the design before writing tests for things that are removed or changed again then.

### DIFF
--- a/src/Middleware/RequestAuthorizationMiddleware.php
+++ b/src/Middleware/RequestAuthorizationMiddleware.php
@@ -44,7 +44,7 @@ class RequestAuthorizationMiddleware
     protected $_defaultConfig = [
         'authorizationAttribute' => 'authorization',
         'identityAttribute' => 'identity',
-        'canMethod' => 'access'
+        'method' => 'access'
     ];
 
     /**
@@ -92,7 +92,7 @@ class RequestAuthorizationMiddleware
         $service = $this->getServiceFromRequest($request);
         $identity = $request->getAttribute($this->getConfig('identityAttribute'));
 
-        if (!$service->can($identity, $this->getConfig('canMethod'), $request)) {
+        if (!$service->can($identity, $this->getConfig('method'), $request)) {
             throw new ForbiddenException();
         }
 
--- a/src/Middleware/RequestAuthorizationMiddleware.php
+++ b/src/Middleware/RequestAuthorizationMiddleware.php
@@ -56,7 +56,7 @@ class RequestAuthorizationMiddleware {
     /**
      * Constructor
      *
-     * @var array $config Configuration options
+     * @param array $config Configuration options
      */
     public function __construct($config = [])
     {
@@ -138,7 +138,7 @@ class RequestAuthorizationMiddleware {
     /**
      * Handles the case no identity is present or it is unauthorized
      *
-     * @param \Authorization\IdentityInterface;
+     * @param \Authorization\IdentityInterface $identity Identity
      * @param \Psr\Http\Message\ServerRequestInterface $request Server request.
      * @param \Psr\Http\Message\ResponseInterface $response Response.
      * @return \Psr\Http\Message\ResponseInterface $response Response.
--- a/src/Middleware/RequestAuthorizationMiddleware.php
+++ b/src/Middleware/RequestAuthorizationMiddleware.php
@@ -16,10 +16,7 @@ namespace Authorization\Middleware;
 
 use Authorization\AuthorizationServiceInterface;
 use Authorization\Exception\ForbiddenException;
-use Authorization\Exception\MissingIdentityException;
-use Authorization\IdentityInterface;
 use Cake\Core\InstanceConfigTrait;
-use Cake\Routing\Router;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
@@ -47,10 +44,6 @@ class RequestAuthorizationMiddleware
     protected $_defaultConfig = [
         'authorizationAttribute' => 'authorization',
         'identityAttribute' => 'identity',
-        'redirectUrl' => null,
-        'unauthorizedHandler' => null,
-        'missingIdentityHandler' => null,
-        'allowAccessWhenIdentityIsMissing' => false,
         'canMethod' => 'access'
     ];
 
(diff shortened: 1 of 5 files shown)


markstory:
> This middleware is useful when you want to authorize your requests, for example each controller and action, against a role based access system or any other kind of authorization process that controls access to certain actions.

Where is that authorization logic going to be placed? Couldn't controller & action pairs also map to actions being done on entities/queries?

burzum:
@markstory in a policy? I've written my own policy resolver and policy for what I want to do.

Why are you so fixed on doing all the authentication stuff based on entities? 😟 If we have a role that is not allowed to even view a listing of something I don't want to go to the controller at all if possible. We have some sections of the app that are completely restricted to certain roles. I don't even see a need to use an entity for this. Why would I have to instantiate the controller and entity first when I simply don't want role X to access Y. Another edge case: What if I want to restrict a certain role from a static page? Do I construct a special entity for that, that doesn't belong to a repository object and instantiate just that in a special method call in the pages controller just to limit /pages/private/* to a certain role?

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+     * @return ResponseInterface A response.
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        $service = $this->getServiceFromRequest($request);
+        $identity = $request->getAttribute($this->getConfig('identityAttribute'));
+
+        if ($identity instanceof IdentityInterface) {
+            if (!$service->can($identity, $this->getConfig('canMethod'), $request)) {
+                $response = $this->handleUnauthorized($identity, $request, $response);
```
robertpustulka: I'd handle redirects using unauthorized handlers mechanism. Just throw `Authorization\Exception\ForbiddenException`.

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+            }
+        } else {
+            $response = $this->handleMissingIdentity($request, $response);
```
robertpustulka: As above - throw `Authorization\Exception\MissingIdentityException` here and let the handler do the rest.

burzum:
@markstory the resource is the parsed request information: Controller, Action, Prefix, Plugin. This is what I want to check against. I have a map of `My/Plugin.Admin/Controller/Action' to associated roles that is checked it's plain simple. Tell me how I could do this even more simple with entities...? I need just a single method and single policy for what I'm doing, the way I do it. I doubt this becomes simpler with entities.

> Also I'm noty sure delegating all authorization logic to the access policy method is a good idea. Is there any other way we could determine the action taken on a request (method, url, something else)?

You suggest this on Slack. ;) As I explained above, I'm mapping the string containing the parts I need to roles. Sure I could instantiate a new entity from the entity class and pass the entity to the policy just for what...? I would take the data again, construct the string identifier and map it again to roles in one method... The point is that I don't want to have a few hundred methods in the policy to write by hand just to check a damn role against a request.

robertpustulka:
> You suggest this on Slack. ;)

This was my suggestion for your case. I'm wondering how can we do it better in a more flexible way so more people could use the request authorization.

markstory:
> The point is that I don't want to have a few hundred methods in the policy to write by hand just to check a damn role against a request. I strongly disagree with the design idea that I must create a can* method for every single thing. When I hook up the authorization to a DB system that manages them I really don't want to create can* methods for every new action! This defeats the whole point of having authorization manageable through an UI and DB.

This is starting to sound like a separate plugin to me. One that manages authorization for controller actions based on database records. Having both in one plugin sounds like it will compromise the clarity of both approaches by having the other approach along for the ride.

burzum:
The DB should not be a hard requirement, ini or php array works as well. I thought this plugin here was thought to be a *generic* foundation and not just something tied to entities. Honestly I'm a little disappointed this has turned out as a "entity first" implementation. 😞

ADmad:
> ..nor we should advertise it as a drop in replacement.

I don't think the new auth plugins were ever meant to be drop in replacements for core's AuthComponent and never advertised as such.

burzum:
@ADmad my point is that the simplicity of a simple implementation is gone.

Laravels policies allow a lot more flexibility out of the box by the way.

* Authorizing Actions Using Policies
  * Via The User Model
  * Via Middleware
  * Via Controller Helpers
  * Via Blade Templates

https://laravel.com/docs/5.6/authorization

markstory:
> So how does this plugin replace the former functionality of the auth components allow() and denie()?

You can use the `skipAuthorization()` method on the component to 'allow', and deny is the default mode.

> I can't recommend people this plugin as a valid replacement if it doesn't provide a way for this nor we should advertise it as a drop in replacement.

Right now it doesn't. But the AuthComponent approach required a controller to be created which was something you were trying to avoid with this implementation. If you're ok with creating the controller instance, I think we could come up with a component based implementation that lets you define an isAuthorized 'hook' on the controller where custom logic can be defined.

> Laravels policies allow a lot more flexibility out of the box by the way.

From their documentation all those integration points are focused on record (entity) operations.

burzum:
@robertpustulka I've already implemented it (less than 200 lines), but using the request for the reasons I've explained a few comments before. Also this code doesn't require a can* method per action, I don't see a reason why it is needed because the actual lookup if this can be accessed is delegated to a method that pulls the required info from config (or DB) and then checks it against the current identity. Also from a performance perspective, why should I instantiate a controller if I'm not even allowed to access it?

burzum:
@robertpustulka @markstory any further comment what needs to be done to finish this?

markstory:
It would be good to have documentation that explains when/why/how to use this class. Right now it is not clear on why/when someone should use this over the policy system we have now.

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param \Psr\Http\Message\ResponseInterface $response Response.
+     * @param callable $next The next middleware to call.
+     * @return ResponseInterface A response.
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        $service = $this->getServiceFromRequest($request);
+        $identity = $request->getAttribute($this->getConfig('identityAttribute'));
+
+        if ($identity instanceof IdentityInterface) {
```
robertpustulka: `|| $identity === null` as we support missing identities in policies now.

burzum:
@markstory I've added docs. @robertpustulka any ideas how to improve the tests?

burzum:
@markstory yet another month has passed, what is still missing?

markstory:
@burzum Nothing. I just forgot.

PR #40, opened 2018-04-18
Allow null as identity for policies

burzum:
Policies can handle non-logged in cases then as well. I've had a discussion with @markstory about this and by removing the requirement of an identity object we can deny or allow access for non-logged cases as well by a policy.

### DIFF
--- a/src/AuthorizationServiceInterface.php
+++ b/src/AuthorizationServiceInterface.php
@@ -25,12 +25,12 @@ interface AuthorizationServiceInterface
      * This method is intended to allow your application to build
      * conditional logic around authorization checks.
      *
-     * @param \Authorization\IdentityInterface $user The user to check permissions for.
+     * @param \Authorization\IdentityInterface|null $user The user to check permissions for.
      * @param string $action The action/operation being performed.
      * @param mixed $resource The resource being operated on.
      * @return bool
      */
-    public function can(IdentityInterface $user, $action, $resource);
+    public function can($user, $action, $resource);
 
     /**
      * Apply authorization scope conditions/restrictions.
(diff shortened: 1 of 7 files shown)


markstory:
An example use case would be: unauthenticated visitors can view a gallery if it is marked public, otherwise it requires the user to be in the selected group of users.

src/AuthorizationServiceInterface.php
```diff
@@ -25,12 +25,12 @@
...
      * conditional logic around authorization checks.
      *
-     * @param \Authorization\IdentityInterface $user The user to check permissions for.
+     * @param \Authorization\IdentityInterface|null $user The user to check permissions for.
      * @param string $action The action/operation being performed.
      * @param mixed $resource The resource being operated on.
      * @return bool
      */
-    public function can(IdentityInterface $user, $action, $resource);
+    public function can($user, $action, $resource);
```
dereuromark: Why removing? Can't you just ? nullable it?

burzum: OK, true, wasn't aware of it, need to get used to the new php features for CakePHP :)

dereuromark: Just bear in mind that this plugin then needs to add `>=7.1` in https://github.com/cakephp/authorization/blob/master/composer.json#L11

burzum: @markstory your opinion on that?

dereuromark: ok. for 4.0 then.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -143,7 +143,7 @@ protected function getIdentity(ServerRequestInterface $request)
         $identityAttribute = $this->getConfig('identityAttribute');
         $identity = $request->getAttribute($identityAttribute);
         if ($identity === null) {
-            throw new MissingIdentityException([$identityAttribute]);
+            return $identity;
```
robertpustulka: Oh no. This makes unauthorized handlers useless and breaks implementations where missing identifiers should be redirected.

Please revert that change.

robertpustulka:
Too bad I have not had a chance to see that before being merged. I don’t agree with this change. How would you redirect missing identities now?

robertpustulka:
@burzum can you update the docs to reflect this change?

PR #43, opened 2018-04-23
Fix remaining deprecations.

markstory:
A few were missed. These need to be fixed to address remaining warnings.


composer.json
```diff
@@ -9,14 +9,14 @@
 	],
 	"type": "cakephp-plugin",
 	"require": {
-		"cakephp/core": "~3.5",
+		"cakephp/core": "^3.6.0",
```
robertpustulka: Do we need to make this +3.6 only plugin?

markstory: We do if we want to use `getRepository()` to avoid the deprecation warning for `repository()`.

dereuromark: Either silence or if/else switch based on version.

markstory:
I've replaced the changes in this pull request with disabling the deprecations. This will allow the plugin to stay 3.5.0+ instead of requiring 3.6.

Issue #42, opened 2018-04-22
AuthorizationComponent::authorize() should now throw an exception


burzum:
I think this should return bool instead. The exception should be thrown in authorizeAction() because this is the automated action check and it is likely that I don't want to display it at all. But authorize() should give me the option to decide what I want to do instead of requiring me to wrap the call in a try/catch.

@markstory told me this on Slack before he went offline:

> The middleware lets you configure how unauthorized exceptions are handled. Does it need to be in the component?

If the article is not allowed to be viewed I want to show a flash message and redirect back to from where ever the user came from and not relying on a generic handler, probably redirecting me to a login page. Or what if I would like to stay on the current page but render a paywall like page? I would have to handle the exception again. An exception feels wrong here in any case. It's not an exception but an expected result that I either can or can't access the action.

markstory:
I'm not a fan of introducing a breaking change to the existing method. What do you think about adding another method that returns `bool` instead of a throwing an exception when authorization fails?

robertpustulka:
I think we could add a `can` method to the component for such cases.

```php
if ($this->Authorization->can('edit', $article)) {
    ...
}
```
^ for authorization checks

```php
$this->Authorization->authorize('edit', $article));
```
^ for restricting access using an exception

---

> I'm not a fan of introducing a breaking change to the existing method.

We already did in #40, #41

burzum:
@markstory  this is what we still have in the readme the the version is still beta. I don't see a problem in making breaking changes until we get pass the RC releases?

> This is still under active development and APIs may change without notice ⚠️

I would prefer one method, but OK. I can add `can` or `isAuthorzed`. Which one do you prefer?

robertpustulka:
>  I don't see a problem in making breaking changes until we get pass the RC releases?

I see because I would need to change my code. As I need to do with changes made in #40.

It's nice though that you are discussing those changes in a public issue before posting and merging a PR.

burzum:
@robertpustulka well, it is still a beta and the readme tells you that breaking changes are expected. I really dislike the idea of already adding legacy code and workarounds to something that is clearly considered as "work in progress". We should get it as good and right for all cases as possible until we are past the RC phase. And the change in 40 was discussed with @markstory on Slack before.

robertpustulka:
> And the change in 40 was discussed with @markstory on Slack before.

On slack, but not in public. The discussion and the decision has been made behind the back of the community. And that's not how the OSS work.

[...]

PR #45, opened 2018-04-23
Adding a can() method to AuthorizationComponent

burzum:
Refs #42

### DIFF
--- a/src/Controller/Component/AuthorizationComponent.php
+++ b/src/Controller/Component/AuthorizationComponent.php
@@ -60,6 +60,30 @@ class AuthorizationComponent extends Component
      */
     public function authorize($resource, $action = null)
     {
+        if ($action === null) {
+            $request = $this->getController()->request;
+            $action = $this->getDefaultAction($request);
+        }
+
+        if ($this->can($resource, $action)) {
+            return;
+        }
+
+        throw new ForbiddenException([$action, get_class($resource)]);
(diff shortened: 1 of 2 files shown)


src/Controller/Component/AuthorizationComponent.php
```diff
@@ -59,6 +59,30 @@ class AuthorizationComponent extends Component
      * @throws \Authorization\Exception\ForbiddenException when policy check fails.
      */
     public function authorize($resource, $action = null)
+    {
+        if ($action === null) {
+            $request = $this->getController()->request;
```
ADmad: Should use ->getRequest(), ->request is deprecated.

burzum: This plugin is still using 3.5?

ADmad: IMO those willing to put in the effort to upgrade there existing projects to using the plugin would also be upgrading to latest cake version too.

burzum: I would like to go to 3.6 soon but we have a ton of plugins that need to upgrade as well. :(

markstory: I think we should stay 3.5+ for a while longer.

robertpustulka:
Docs need to be updated with the new method.

Issue #60, opened 2018-10-30
Component doesn't handle non-object resources


Zuluru:
In a number of places through the plugin, calls to `get_class($resource)` are preceded by `if (is_object($resource))`. But in `AuthorizationComponent::authorize`, if the action is not authorized, the parameters passed to the `ForbiddenException` constructor don't have this check, they just call `get_class($resource)`. I am building a resolver that just takes a string for the policy class name instead of deducing it from an object, and it fails in these cases because of this. It seems that this is likely just an oversight; I can easily put together a PR to correct it, if it's an actual defect instead of intentional behaviour.

markstory:
> I can easily put together a PR to correct it, if it's an actual defect instead of intentional behaviour.

Supporting resolvers with non-object arguments is something we'd like to do. We have clearly not done a good job of testing that behavior in an integrated way.

Zuluru:
I can submit my string-based resolver for consideration at some point, if you want, but it's not quite ready to see the light of day.

markstory:
> I don't see an obvious way to add a test for the fix without adding an entire string-based resolver class.

That's ok. It could live in the `tests` directory and won't be part of the autoloader for the plugin.

markstory:
Hoping to avoid adding stub class to the tests folder? We already have a bunch of [stub classes](https://github.com/cakephp/authorization/tree/master/tests/test_app/TestApp).

dereuromark:
Resolved with opened PR.

PR #61, opened 2018-10-31
Allow for non-object resources in authorize

Zuluru:
Addresses issue #60.

### DIFF
--- a/src/Controller/Component/AuthorizationComponent.php
+++ b/src/Controller/Component/AuthorizationComponent.php
@@ -71,7 +71,7 @@ class AuthorizationComponent extends Component
 
         if (is_object($resource)) {
             $name = get_class($resource);
-        } else if (is_string($resource)) {
+        } elseif (is_string($resource)) {
             $name = $resource;
         } else {
             $name = gettype($resource);
--- a/src/Controller/Component/AuthorizationComponent.php
+++ b/src/Controller/Component/AuthorizationComponent.php
@@ -69,7 +69,7 @@ class AuthorizationComponent extends Component
             return;
         }
 
-        throw new ForbiddenException([$action, get_class($resource)]);
+        throw new ForbiddenException([$action, is_object($resource) ? get_class($resource) : (is_string($resource) ? $resource : gettype($resource))]);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/test_app/TestApp/Policy/StringResolver.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Very simple policy resolver that accepts string policy names.
+ */
+class StringResolver implements ResolverInterface
+{
+    /**
+     * Get a policy for a string.
+     *
+     * @param string $resource The resource.
+     * @return object
```
dereuromark: According to used class (constructor of it), this is `\Authorization\Policy\ResolverInterface`

Zuluru: The @return doc? That was copied directly from the OrmResolver.

Zuluru: Being stricter seems like a good idea, but is outside the scope of this PR, I think. Addresses a different issue.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -69,7 +69,7 @@ public function authorize($resource, $action = null)
             return;
         }
 
-        throw new ForbiddenException([$action, get_class($resource)]);
+        throw new ForbiddenException([$action, is_object($resource) ? get_class($resource) : (is_string($resource) ? $resource : gettype($resource))]);
```
markstory: I would avoid nested ternary operators in PHP they are a bit funny.

tests/test_app/TestApp/Policy/StringResolver.php
markstory: I don't think Policies have a common interface though.

Zuluru: I think he was suggesting that they should? Which, like I said, would be a separate issue.

markstory: Yes it is a separate conversation.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -69,7 +69,14 @@ public function authorize($resource, $action = null)
             return;
         }
 
-        throw new ForbiddenException([$action, get_class($resource)]);
+        if (is_object($resource)) {
+            $name = get_class($resource);
+        } else if (is_string($resource)) {
```
dereuromark: elseif

markstory:
Thank you 👍

Issue #62, opened 2018-11-22
Flash messages and redirect URLs for authorization failures


Zuluru:
I don't see any way to set the flash message that gets used in case of authorization (or authentication) failures, or provide custom redirect URLs for different policy failures (e.g. redirect to a view when edit isn't allowed). I've tried added this capability in my app by writing my own "unauthorized" handler and throwing custom exceptions from my policies; the handler pulls the required details from these and proceeds accordingly. But while the exceptions work okay for actual authorization checks (`$this->Authorization->authorize()`), they cause problems when doing authorization *tests* (`$this->Authorization->can()`).

I've had to write quite a bit of code to get close to functionality that the old Auth component had, so I feel like I may be missing something. Or maybe the plugins just haven't had it added yet? Happy to work towards a PR if this is the sort of thing that the plugin should support natively.

markstory:
> I don't see any way to set the flash message that gets used in case of authorization (or authentication) failures, or provide custom redirect URLs for different policy failures (e.g. redirect to a view when edit isn't allowed).

This isn't something we've implemented. Custom exceptions is how I would approach this problem as well. What isn't working well with `can()` and custom exceptions? Perhaps we can make that better.

Zuluru:
I have a number of functions that I've built into my test suite, for checking return values (status codes, flash messages, redirect URLs, body content) in various test scenarios. I'm working my way though reducing this list down to the bare minimum of unique authentication and authorization scenarios that I need to deal with. (I suspect that I may be trying overly hard to give helpful messages and redirects to people that try to do things they aren't allowed to do, but between cached search engine results and emailed links that go stale, there are a number of situations where someone *could* have done something, but no longer can, and I want to be user-friendly in such cases, to reduce the number of support emails if nothing else.) Anyway, this winnowing process is taking some time. Hope to have some results from that, and an accompanying analysis of what works and what doesn't soon.

In the meantime, I did come across something (for probably the third time) that is part of this. The authorization component can be given a 'skipAuthorization' list. But all that happens with these functions is that they are not flagged as not having been authorized by the end of the process. If one calls `$this->Authorization->authorize()` in one of those functions, the authorization is still run. I can see arguments both for and against this behaviour; I just wanted to confirm that it is intentional as it stands, not a bug that will be fixed at some point.

Zuluru:
[...]

PR #63, opened 2018-12-05
Catch certain exceptions in the component's can function and return false instead

Zuluru:
This addresses part of issue #62. Note that the *service*'s `can` function has not been changed, only the *component*'s. We do want exceptions thrown from policies to get through the `authorize` function. If we want to change the service's `can` function to catch exceptions, then we might create a new `authorize` function in there as well, which the component can call, or add a parameter indicating whether or not exceptions should be caught.

### DIFF
--- a/src/Controller/Component/AuthorizationComponent.php
+++ b/src/Controller/Component/AuthorizationComponent.php
@@ -65,7 +65,7 @@ class AuthorizationComponent extends Component
             $action = $this->getDefaultAction($request);
         }
 
-        if ($this->can($resource, $action)) {
+        if ($this->_can($resource, $action)) {
             return;
         }
 
@@ -80,7 +80,13 @@ class AuthorizationComponent extends Component
     }
 
     /**
-     * Check the policy for $resource, returns true if the action is allowed
+     * Check the policy for $resource, returns true if the action is allowed.
+     *
+     * To support policies throwing their own exceptions (potentially required
+     * in order for `authorize` checks to work in advanced scenarios), any
+     * Forbidden and MissingIdentity exceptions will be caught and result in
+     * a false return value. If you have your own custom exceptions, you can
+     * ensure that they will be dealt with here by extending one of these two.
      *
      * If $action is left undefined, the current controller action will
      * be used.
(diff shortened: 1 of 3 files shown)


src/Controller/Component/AuthorizationComponent.php
```diff
@@ -91,11 +97,30 @@ public function authorize($resource, $action = null)
...
+    /**
+     * Check the policy for $resource, returns true if the action is allowed
+     *
+     * @param object $resource The resource to check authorization on.
+     * @param string $action The action to check authorization for.
+     * @return bool
+     */
+    protected function _can($resource, $action)
+    {
+        $request = $this->getController()->request;
```
markstory: Should line 125 also use this request?

Zuluru:
The more I think about it, the more I think we should add an `authorize` method in the service. The asymmetric behaviour of the two `can` functions (one catching exceptions, one allowing them) bothers me.

markstory:
>  The asymmetric behaviour of the two can functions (one catching exceptions, one allowing them) bothers me.

Should we consider a new `can()` that is 'safe'?

Zuluru:
That feels a bit Rube Goldberg to me.

We have a `can` in the service. Given that it's not currently catching exceptions, it seems unlikely to me that there's much in the way of existing policies in the wild that are throwing exceptions. And if they are, they must be catching them manually, or else things would be breaking for them, if my experience is any indication. So, the potential impact of making that `can` catch exceptions should be very minimal; again, my experience was that whenever I catch anything, I have to do exactly the same as I do if it returns false, and thus if it starts returning false I just have some dead code that's no longer being hit, but behaviour is identical.

We (or, at least, I) do need something in the service that passes all exceptions through. But AFAICT that's only needed by the `authorize` function in the component. What would be the point of a `can` function at the component level that lets all exceptions through? That's precisely the `authorize` function.

So, if we need a function in the component that passes exceptions, and one that doesn't, then those roles can be filled by the existing `authorize` and `can` respectively, no additions needed. And if we want those same things in the service, then it seems that having functions with the same names and semantics is the way to go.

@jeremyharris, any thoughts on this?

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -91,13 +97,32 @@ public function authorize($resource, $action = null)
...
     {
-        $request = $this->getController()->request;
         if ($action === null) {
+            $request = $this->getController()->request;
             $action = $this->getDefaultAction($request);
         }
 
+        try {
+            return $this->_can($resource, $action);
+        } catch (ForbiddenException $ex) {
```
robertpustulka: Why would `can` even throw one of these exceptions? `can` should return only `bool`, this isn;t supposed to throw any authorization related exceptions.

Zuluru: @markstory, we've had a little conversation over there already. As yet unresolved.

robertpustulka:
Does #64 solve the issue, can we close this?

Zuluru:
It probably solves it, though I expect it'll be a couple weeks before I will find time to try it out. At the very least, if it doesn't fully solve it, the solution for anything left is not likely to look much like this PR.

PR #64, opened 2018-12-21
Add policy check Result object

robertpustulka:
Refs discussions: https://github.com/cakephp/authorization/issues/62 https://github.com/cakephp/authorization/pull/63 

This allows policies to return result objects that wrap status boolean and a reason string. Policies can still return just bools so this change should be transparent to those who don't use the new result object.

No tests added yet, as I'm not 100% sure this is a good solution to the problems pointed out in issues linked above.

### DIFF
--- a/src/Exception/ForbiddenException.php
+++ b/src/Exception/ForbiddenException.php
@@ -15,6 +15,8 @@
 
 namespace Authorization\Exception;
 
+use Authorization\Policy\ResultInterface;
+
 class ForbiddenException extends Exception
 {
     /**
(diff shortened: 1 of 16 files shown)


src/Exception/ForbiddenException.php
```diff
@@ -26,4 +28,42 @@ class ForbiddenException extends Exception
...
+     * @param \Exception|null $previous the previous exception.
+     */
+    public function __construct($result = null, $message = '', $code = null, $previous = null)
+    {
+        if ($result instanceof ResultInterface) {
+            $this->result = $result;
+
+            parent::__construct($message, $code, $previous);
+        } else {
+            //shift off first argument fort BC
```
markstory:
```suggestion
            //shift off first argument for BC
```

markstory:
Looks good to me once we have some tests. The reason + status should cover the use cases we talked about.

robertpustulka:
Tests added.

@markstory should we add an option to the service to convert all bools returned from policies to results? This could be disabled by default, but when enabled it would make working with `can` return values more consistent as you'd expect the method to always return `ResultInterface` even when your policy returns `bool` internally.

markstory:
Would that option change the return value of `can()`?

robertpustulka:
Yes, that’s the idea. With the option enabled you wouldn’t need to remember which of your policies return bool and which an object. You expect objects to be returned everywhere.

markstory:
Do you think result objects will be inconsistently used by an application that we need to wrap results?

I would hope that applications use result objects consistently or that they are aware enough to manage the additional complexity they require.

robertpustulka:
> I would hope that applications use result objects consistently or that they are aware enough to manage the additional complexity they require.

Yes, I think you're right. 

OK. So this should be good to go if the code is fine.

PR #68, opened 2019-01-29
Add convenience policy interface.

dereuromark:
Improve docs and add a convenience policy interface to implement.
Saves a lot of time here and clarifies the usage.

### DIFF
--- a/src/Policy/RequestPolicyInterface.php
+++ b/src/Policy/RequestPolicyInterface.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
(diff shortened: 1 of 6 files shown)


dereuromark:
@markstory The doc fixes should also help to make it understand easier what variable is what target argument on objects. $map itself doesnt say much and doesnt "fit" to any of the objects.
I was confused here too for a minute until I was looking more deeply into the return and argument typehints and figured it out.

src/Policy/RequestPolicyInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+namespace Authorization\Policy;
+
+use Authorization\IdentityInterface;
+use Cake\Http\ServerRequest;
+
+/**
+ * This interface should be implemented by your request policy class.
+ */
+interface RequestPolicyInterface
```
markstory: We don't check against this interface anywhere. Is there a reason to add it?

PR #69, opened 2019-01-29
Enable php73 check

dereuromark:
and use reasonable test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ language: php
 php:
   - 5.6
   - 7.0
-  - 7.1
-  - 7.2
+  - 7.3
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,8 +3,7 @@ language: php
 php:
   - 5.6
   - 7.0
-  - 7.1
-  - 7.2
+  - 7.3
```
markstory: I would prefer to keep 7.2 over 7.0.

dereuromark: why? we need the upper and lower limits of the language per major.
rest in between should be fine based on this.

markstory: In theory yes, in practice we've found a bunch of problems where type hinting works differently between 7.1 and 7.2.

dereuromark: But it would be the same for 7.3 if it changes after another minor. The changes are always linear here, not fluctuating.
7.0 => 7.1 => 7.2 => 7.3
Whatever you see happening between 7.1 and 7.2 is also present in 7.3.
Can you show be one example that would not be true for the case as I outline it?

PR #73, opened 2019-02-13
Add convenience methods for common config keys.

robertpustulka:
This adds convenience methods for `actionMap` and `authorizeModel` configuration keys.


markstory:
Nice work :+1"

PR #77, opened 2019-03-18
Update plugin to use phauthentic/authorization as dependency.

ADmad:
~~Temporarily using my fork of `phauthentic/authorization` until https://github.com/Phauthentic/authorization/pull/2 gets merged.~~


ADmad:
> [Composer\Repository\InvalidRepositoryException]                             
  No valid composer.json was found in any branch or tag of https://github.com  
  /ADmad/authorization, could not load a package from it.

It works fine for me locally :angry:.

docs/Checking-Authorization.md
```diff
@@ -9,17 +9,6 @@ authorization. The middleware will wrap your request `identity` with an
...
-if ($user->can('delete', $article)) {
-    // Do delete operation
-}
-```
-
-If your policies return [Result objects](../Policies.md#policy-result-objects) 
-be sure to check their status as  `can()` returns the result instance:
-
-```php
-// Assuming our policy returns a result.
```
markstory: Why remove this example?

docs/Component.md
```diff
@@ -70,7 +70,7 @@ when permission is denied. If you want to check authorization and get a boolean
 result you can use the `can()` method:
 
 ```php
-if ($this->Authorization->can($article, 'update')) {
+if ($this->Authorization->can($article, 'update')->getStatus()) {
```
markstory: In the paragraph above it mentions that if you want a boolean you can use `can()`. Is that no longer the case?

docs/Policies.md
```diff
@@ -100,14 +100,15 @@ in your policy:
...
+use Phauthentic\Authorization\Policy\Result;
+use Phauthentic\Authorization\Policy\BeforePolicyInterface;
 
 class ArticlesPolicy implements BeforePolicyInterface
 {
     public function before($user, $resource, $action)
     {
         if ($user->getOriginalData()->is_admin) {
-            return true;
+            return new Result(true);
```
markstory: Is there no longer support for returning bare booleans?

ADmad: Yes support for boolean return will be removed in next major. Instance of `ResultInterface` must always be returned now. This allows for cleaner API and return type hinting.

I'll fix the docs wherever it still mentions boolean return.

markstory: Doesn't always requiring Result objects make both building Policies and using can() harder more awkward though?

Could we add methods like `Result::success()` and `Result::failure()` as porcelain methods for making pass/fail result objects?

ADmad: You mean static methods?

markstory: Yes. Having to wrap booleans is a lot of syntax. I'd prefer to not break that interface as it doesn't benefit the end user at all. I understand it makes typehinting easier, but its at the detriment of usability.

ADmad: I agree that having to create an object instead of returning boolean when one doesn't need to set a custom reason (which would be majority of time) is not user friendly.

We could continue allowing policy classes to return boolean or result object.

But for `IdentifyInterface::can()` and `AuthorizationInterface::can()` always returning objects is better as then one doesn't have to check return value type in app code. Implementation of these method could handle creating of result object from booleans if required.

/cc @robertpustulka @burzum

robertpustulka: AFAIR policy classes were not required to return objects. I agree that’s not necessary. Identity and service should return objects only for consistency though.

markstory: Instead of forcing result objects to be used in can() should we consider exposing another method that returns result objects?

My concern is that we will be breaking the core methods of this plugin and adding complexity everywhere when it is rarely necessary. Having can() return bool|Result was a mistake in hindsight. We could correct that erring on the side of simplicity instead of adding complexity like we would be doing with this change.

ADmad: I have made a PR https://github.com/Phauthentic/authorization/pull/4 to allow policy method to return boolean again.

ADmad: > My concern is that we will be breaking the core methods of this plugin and adding complexity everywhere when it is rarely necessary.

Isn't worrying about breaking methods for the plugin which had it's 1st stable release (prematurely IMO) only few months ago being overly conservative?

This discussion also makes me second guess whether using `phauthetic/authorization` as a dependency is a good idea. The goals of this plugin and the agnostic lib could diverge in future causing more friction with regards to the API.

My initial goal was to have the plugin ready for Cake 4 which has already been achieve with #76.

markstory: Even if we were pre 1.0 I wouldn't like the change to adopt result objects everywhere. I feel it adds complexity and overhead where it isn't useful in the majority of scenarios.

ADmad: I have no interest in continuing this discussion and am fine simply closing this PR.

As I mentioned 2.x branch is already Cake 4 ready.

PR #83, opened 2019-05-12
Run only for PHP 7.2+

dereuromark:


markstory:
I will get ci passing again tonight.

PR #86, opened 2019-05-13
Always return ResultInterface instance from AuthenticationService::can().

ADmad:
This allows for cleaner app code as it avoids checking type of the returned value.
The convenience of returning boolean from policy classes is still retained.

### DIFF
--- a/src/AuthorizationService.php
+++ b/src/AuthorizationService.php
@@ -51,48 +51,67 @@ class AuthorizationService implements AuthorizationServiceInterface
     /**
      * @inheritDoc
      */
-    public function can(?IdentityInterface $user, string $action, $resource): ResultInterface
+    public function can(?IdentityInterface $user, string $action, $resource): bool
+    {
+        $result = $this->performCheck($user, $action, $resource);
+
+        return is_bool($result) ? $result : $result->getStatus();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function canResult(?IdentityInterface $user, string $action, $resource): ResultInterface
+    {
+        $result = $this->performCheck($user, $action, $resource);
+
+        return is_bool($result) ? new Result($result) : $result;
+    }
+
+    /**
+     * Check whether the provided user can perform an action on a resource.
+     *
+     * @param \Authorization\IdentityInterface|null $user The user to check permissions for.
+     * @param string $action The action/operation being performed.
+     * @param mixed $resource The resource being operated on.
+     * @return bool|\Authorization\Policy\ResultInterface
+     */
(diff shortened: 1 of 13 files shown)


src/AuthorizationService.php
```diff
@@ -50,34 +51,48 @@ public function __construct(ResolverInterface $resolver)
     /**
      * @inheritDoc
      */
-    public function can(?IdentityInterface $user, string $action, $resource)
+    public function can(?IdentityInterface $user, string $action, $resource): ResultInterface
```
markstory: If the service is consistently returning a Result how does `Identity::can()` return booleans?

ADmad: I forgot to update Identity class :) Guess we don't have any tests which cover integration between both service and identity.

ADmad: So there was a test but it only tested with the service returning Result hence nothing failed :).

markstory: That's unfortunate

markstory: My question was more around the breakage and complexity spillage that applications are exposed to. By always returning a Result we force everyone to deal with the most complex scenario (a Result) when the primary usage is returning booleans.

ADmad: It would add one extra method call for checks but it would make your app code future proof. Boolean results might suffice when starting out but let's say down the line you want to add / update a policy which returns `Result` instance, since you want to provide a message with reason for failure.

You would then have to go about updating all checks in your app to check whether return type of `can()` is boolean or `Result` instance.

I am open to alternate suggestions to avoid such code refactoring.

markstory: Could we add a `canResult()` method to the IdentityDecorator & AuthorizationService? That would let people opt-in to 'expert mode' of getting results back.

ADmad: We could, and change `can()` to always return boolean?

markstory: Yeah. It is a breaking change from 1.x but a smaller one than having can() always return Result objects.

ADmad:
@markstory Changes done as suggested.

src/AuthorizationService.php
```diff
@@ -50,34 +51,67 @@ public function __construct(ResolverInterface $resolver)
...
+     */
+    protected function resultTypeCheck($result)
+    {
+        if (is_bool($result) || $result instanceof ResultInterface) {
             return $result;
         }
 
-        return $result === true;
+        throw new Exception(sprintf(
+            'Pre-authorization check must return `%s`, `bool` or `null`.',
```
markstory: Nice error message :+1:

ADmad:
New method `canResult()` was added which return `ResultInterface` instance.

PR #87, opened 2019-05-13
Update OrmResolver.php

dangerdan:
Fixed deprecated repository() call.

### DIFF
--- a/src/Policy/OrmResolver.php
+++ b/src/Policy/OrmResolver.php
@@ -69,7 +69,7 @@ class OrmResolver implements ResolverInterface
             return $this->getRepositoryPolicy($resource);
         }
         if ($resource instanceof QueryInterface) {
-            return $this->getRepositoryPolicy($resource->repository());
+            return $this->getRepositoryPolicy($resource->getRepository());
         }
         $name = is_object($resource) ? get_class($resource) : gettype($resource);
         throw new MissingPolicyException([$name]);
--- a/src/Policy/OrmResolver.php
+++ b/src/Policy/OrmResolver.php
@@ -69,7 +69,8 @@ class OrmResolver implements ResolverInterface
             return $this->getRepositoryPolicy($resource);
         }
         if ($resource instanceof QueryInterface) {
-            return $this->getRepositoryPolicy($resource->getRepository());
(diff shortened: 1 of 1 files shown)


src/Policy/OrmResolver.php
```diff
@@ -69,7 +69,7 @@ public function getPolicy($resource)
             return $this->getRepositoryPolicy($resource);
         }
         if ($resource instanceof QueryInterface) {
-            return $this->getRepositoryPolicy($resource->repository());
+            return $this->getRepositoryPolicy($resource->getRepository());
```
saeideng: `getRepository()` does not exist in cake3.5 
first you should check for availability

markstory:
This would require bumping the minimum required CakePHP version to at least 3.6 like @saeideng mentioned.

dangerdan:
I've updated this now with a check for availability

src/Policy/OrmResolver.php
dereuromark: Instead of the antipattern of method joggling/testing, I would rather raise the limit to 3.6+ at this point then.

src/Policy/OrmResolver.php
```diff
@@ -69,7 +69,8 @@ public function getPolicy($resource)
             return $this->getRepositoryPolicy($resource);
         }
         if ($resource instanceof QueryInterface) {
-            return $this->getRepositoryPolicy($resource->repository());
+            $repository = $resource->getRepository() ?? $resource->repository();
```
markstory: This check will only work in PHP7.1+. This plugin currently supports PHP 5.6+. You would need to use `method_exists()` to check for the method before calling it instead of a null-coalesce.

robertpustulka:
Maybe just bump the min CakePHP version to 3.6 and release the plugin as 1.3?

cleptric:
Doesn't feel right to me doing this. While you could probably argue, that rasing the min Cake version in a new minor release is still in sync with smever, we unnecessarily could cause trouble for folks out there that still use 3.5

dereuromark:
@cleptric That is not true. The opposite is, raising is fine here as long as done in a reasonable way which is the case here.

dereuromark:
The minimum is now 3.6
I opened a PR.

PR #92, opened 2019-07-04
Extends authentication identity interface to resolve exception on auth

skie:
On auth in cakephp application happans next error:

Return value of Authentication\Controller\Component\AuthenticationComponent::getIdentity() must implement interface Authentication\IdentityInterface or be null, instance of Authorization\IdentityDecorator returned 

To resolve the best to extends authentication identity interface in authorization level.

### DIFF
--- a/src/IdentityInterface.php
+++ b/src/IdentityInterface.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace Authorization;
 
 use ArrayAccess;
+use Authentication\IdentityInterface as AI;
 
 /**
  * Interface for describing identities that can have authorization checked.
@@ -25,7 +26,7 @@ use ArrayAccess;
  * and uses ArrayAccess to expose public properties of the wrapped identity
  * implementation.
  */
-interface IdentityInterface extends ArrayAccess
+interface IdentityInterface extends ArrayAccess, AI
 {
     /**
      * Check whether the current identity can perform an action.
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     },
     "require-dev": {
         "cakephp/cakephp-codesniffer": "dev-next",
+        "cakephp/authentication": "2.x-dev as 2.0",
         "cakephp/bake": "4.x-dev as 2.0.0",
         "phpunit/phpunit": "^8.0"
     },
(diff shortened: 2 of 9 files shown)


ADmad:
Don't really like making the authentication plugin a hard dependency for authorization plugin. 

Instead of having `Authorization\IdentityInterface` extend `Authentication\IdentityInterface` I would add a new identity class which implements both interfaces. `AuthorizationMiddleware` then check if `Authentication\IdentityInterface` exists and use the new identity class by default instead of existing `Authorization\IdentityDecorator`.

skie:
How would proposed solution remove dependency from Authentication plugin. It still be reffered when in new class. Alternative sollution would be change strict type check in Authentication\Controller\Component\AuthenticationComponent::getIdentity
Or identity could be stored in diferent attributes for Authentication and Authorization

ADmad:
> How would proposed solution remove dependency from Authentication plugin. It still be reffered when in new class.

It would be an optional dependency. The class which implements identity interface of authentication won't be used if authentication plugin is not loaded.

src/IdentityInterface.php
```diff
@@ -17,6 +17,7 @@
 namespace Authorization;
 
 use ArrayAccess;
+use Authentication\IdentityInterface as AI;
```
markstory: This makes authorization depend on authentication, which isn't great. Should we add an IdentityInterface to CakePHP core? That would allow both plugins to leverage some common traits and authorization could layer on `can()` and `applyScope`, or does that just move the problem around?

ADmad:
[Here's](https://github.com/cakephp/authorization/compare/2.x...2.x-identity-class) my proposed solution.

src/IdentityInterface.php
skie: In fact, changing `identity` attribute with another object from `Authorization` already creates implicit dependency on `Authentication`. So create separate attribute (like `guard`, or `authorizer`) which implements `Authorization\IdentityInterface` could be the best option if we want to use `Authentication` and `Authorization` togather. Right now this bug real stopper.

skie:
> [Here's](https://github.com/cakephp/authorization/compare/2.x...2.x-identity-class) my proposed solution.

That's work for me.

PR #94, opened 2019-07-05
Add new Identity class which implements Authentication\IdentityInterface too.

ADmad:
Alternative to #92.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "psr/http-server-middleware": "^1.0"
     },
     "require-dev": {
+        "cakephp/authentication": "2.x-dev",
         "cakephp/cakephp-codesniffer": "dev-next",
         "cakephp/bake": "4.x-dev as 2.0.0",
         "phpunit/phpunit": "^8.0"
--- a/src/Identity.php
+++ b/src/Identity.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         2.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -17,6 +17,7 @@
         "psr/http-server-middleware": "^1.0"
     },
     "require-dev": {
+        "cakephp/authentication": "2.x-dev",
```
markstory: Should this also be in `suggest` for if someone wants to use the `Identity` class in this plugin?

ADmad: I think only someone already using authentication would be interested in this class.

The new class doesn't provide anything which would make someone chose this class first and then make them use authen. lib.

src/Identity.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * An identity decorator implementing both Authorization\IdentityInterface and
+ * Authentication\IdentityInterface.
+ */
+namespace Authorization;
+
+use Authentication\IdentityInterface as AuthenIdentityInterface;
+
+class Identity extends IdentityDecorator implements AuthenIdentityInterface
```
markstory: Should the docs be updated to note that this class will used by default if the other plugin is installed?

ADmad: Will do.

ADmad: Doc updated.

src/Identity.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @return string|int|null
+     */
+    public function getIdentifier()
+    {
+        if ($this->identity instanceof AuthenIdentityInterface) {
+            return $this->identity->getIdentifier();
+        }
+
+        return $this->identity['id'];
```
robertpustulka: This should probably be configurable.

ADmad: Generally this line will never be executed as the class is only meant to be used when `authentication` plugin is used and hence the `$identity` property should be `Authentication\IdentityInterface` instance. I probably wrote this this way initially to just make running the test suite easier. 

I can change the method to simply `return $this->identity->getIdentifier();` and tests run fine now. Users can always specify their own decorator class for the `AuthorizationMiddleware` if this doesn't work for them.

Keeping the method as it is and allowing `id` to be configurable will requiring changing how `identityDecorator` config can be set. Currently it can just be a string. I don't really feel like going about making changes for it.

robertpustulka: I understand. So maybe throw an exception then?

ADmad: Sounds good, will do.

ADmad: I took a slightly different approach and modified the constructor so that it only takes `Authentication\IdentityInterface` as argument. That avoid any additional checks or exceptions in methods.

PR #95, opened 2019-07-11
fix deprecated repository()

erfunath:


erfunath:
Just saw the other request to do the same thing.

saeideng:
you should not receive deprecation warning , if you receive , problem should fix 
current version of authorization require cake 3.5.x
and `getRepository` added since 3.60
you can check and use this method if `getRepository()` exist

PR #113, opened 2019-12-16
Use stable releases of dependencies

ADmad:


dereuromark:
Problem 1
    - Installation request for cakephp/bake ^2.0 -> satisfiable by cakephp/bake[2.0.0].
    - cakephp/bake 2.0.0 requires wyrihaximus/twig-view dev-master -> satisfiable by wyrihaximus/twig-view[dev-master] but these conflict with your requirements or minimum-stability.

is that were a branch aliases would come in handy?

ADmad:
Missed the fact that this plugin uses bake. Can't remove the "minimum-stability" until bake's dependency twigview get's tagged. So will have a wait a while to merge this.

PR #121, opened 2020-01-30
function expects one parameter

back2back:
`public function getAuthorizationService(ServerRequestInterface $request) : AuthorizationServiceInterface
{
$resolver = new OrmResolver();
return new AuthorizationService($resolver);
}`

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -46,7 +46,7 @@ it starts handling the request. This hook method allows your application to
 define the ``AuthorizationService`` it wants to use. Add the following method your
 **src/Application.php**::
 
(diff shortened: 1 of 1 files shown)


docs/en/index.rst
```diff
@@ -46,7 +46,7 @@ it starts handling the request. This hook method allows your application to
 define the ``AuthorizationService`` it wants to use. Add the following method your
 **src/Application.php**::
 
-    public function getAuthorizationService(ServerRequestInterface $request, ResponseInterface $response)
+    public function getAuthorizationService(ServerRequestInterface $request)
```
ADmad: Return typehint also needs to be added

markstory: I'll fix that up post merge.

PR #123, opened 2020-02-17
Translation for Portuguese - 2.x

tavaresgerson:


markstory:
Thanks for getting this started. I will get it merged as soon as I have time to get deployments updated for this translation.

dereuromark:
Is this really useful/helpful?
These kinf of docs are soon outdated and lying and will cause more harm down the road.
Especially for small plugins like these that have even a way lower maintainer number than the core framework package(s) itself.

markstory:
I think complete translations are helpful. Duplicating the english docs is not as helpful as they will not ever be maintained and may set future translators off with stale information.

ADmad:
Unless all pages are translated I am not in favor of merging this. People's initial gusto soon wears off as they get busy / lost interest and incomplete translations almost always remain incomplete.

tavaresgerson:
I'am compromised translate this documentation, the cakephp and other plugins is part for my jobs and my too collegues. Is not entusiasm

dereuromark:
Given that this is incomplete and no further activity for a year, we can close this.
The topic is difficult to keep up to date anyway, since the primary language (EN) is also constantly updated.

PR #126, opened 2020-03-16
Update can method to canResult

ideodora:
I guess ```can``` method returns boolean. ```canResult``` method will do so.


markstory:
Thanks :+1:

PR #129, opened 2020-03-26
Add canResult to docs.

### LINKED ISSUE #127: method canResult for User entity
in docs is missing that `canResult `method in User entity must be declared

markstory:
Fixes #127


dereuromark:
@markstory Not a big fan of deleting merged branches, are you? ;) Maybe we should activate the auto cleanup also for these plugins.

markstory:
I just forgot. I was trying to quickly get through a bunch of pull requests.

PR #131, opened 2020-04-14
Add unauthorizedRedirect

cleptric:
Quick draft to add an `unauthorizedRedirect` option to the `RequestAuthorizationMiddleware`.
Some open questions:

- Shall we handle a 404 by the `RoutingMiddleware`?
- Is a simple string enough, or do we need `Router` support?
- Shall we append the previous URL to the redirect?

### DIFF
--- a/src/Middleware/RequestAuthorizationMiddleware.php
+++ b/src/Middleware/RequestAuthorizationMiddleware.php
@@ -19,12 +19,12 @@ namespace Authorization\Middleware;
 use Authorization\AuthorizationServiceInterface;
 use Authorization\Exception\ForbiddenException;
 use Cake\Core\InstanceConfigTrait;
+use Laminas\Diactoros\Response\RedirectResponse;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use RuntimeException;
-use Zend\Diactoros\Response\RedirectResponse;
 
 /**
  * Request Authorization Middleware
@@ -99,11 +99,11 @@ class RequestAuthorizationMiddleware implements MiddlewareInterface
         $result = $service->canResult($identity, $this->getConfig('method'), $request);
         if (!$result->getStatus()) {
             $redirectUrl = $this->getConfig('unauthorizedRedirect');
-            if (!empty($redirectUrl)) {
+            if ($redirectUrl) {
                 return new RedirectResponse($redirectUrl);
-            } else {
-                throw new ForbiddenException($result);
             }
+
+            throw new ForbiddenException($result);
         }
 
         return $handler->handle($request);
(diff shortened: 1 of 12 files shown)


markstory (commented):
Seems reasonable to me.

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -96,7 +98,12 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
         $result = $service->canResult($identity, $this->getConfig('method'), $request);
         if (!$result->getStatus()) {
-            throw new ForbiddenException($result);
+            $redirectUrl = $this->getConfig('unauthorizedRedirect');
+            if (!empty($redirectUrl)) {
```
dereuromark, line 102: Cloaking not necessary here, 

    if ($redirectUrl) {

suffices and is identical

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -96,7 +98,12 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+                return new RedirectResponse($redirectUrl);
+            } else {
```
dereuromark, line 104: Else not necessary

dereuromark:
Looks like a good idea, yeah.

src/Middleware/RequestAuthorizationMiddleware.php
```diff
@@ -24,6 +24,7 @@
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use RuntimeException;
+use Zend\Diactoros\Response\RedirectResponse;
```
ADmad, line 27: This should be `Laminas` now not `Zend`. Also it would require adding `laminas/diactoros` as a composer dependency.

dereuromark: If any of those ship with it, we might not need the verbose addition in root composer json here: https://github.com/cakephp/authorization/blob/master/composer.json#L13-L17

ADmad: They don't.

src/Policy/MapResolver.php
```diff
@@ -88,6 +88,8 @@ public function map(string $resourceClass, $policy)
     /**
      * {@inheritDoc}
      *
+     * @param mixed $resource A resource that the access is checked against.
+     * @return mixed
```
dereuromark, line 92: There is no interface?
I think at least `object` is true, though.

dereuromark:
Do we still want to finish this?

othercorey:
Can't this be done using the AuthorizationMiddleware exception handling? That can redirect from a ForbiddenException.

dereuromark:
@cleptric What do you think?

saeideng:
> Can't this be done using the AuthorizationMiddleware exception handling? 

> Is a simple string enough, or do we need Router support?

Router support will available after merging  #192 
so do mentioned PR can cover this PR?

ADmad:
Redirection is now possible through 'unauthorizedHandler` added in #211

PR #151, opened 2020-11-10
Add github actions

othercorey:


othercorey:
Not sure why prefer lowest is failing on GA.

othercorey:
> Not sure why prefer lowest is failing on GA.

We have to add plugin-installer dependencies for composer v2 to ensure 1.3+.

othercorey:
@markstory Can you look at the console integration test failures on windows?

.travis.yml
```diff
@@ -1,50 +0,0 @@
...
-  - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION != 7.3 ]]; then vendor/bin/phpunit; fi
-
-  - if [[ $PHPCS = 1 ]]; then vendor/bin/phpcs -n -p src/ tests/; fi
-  - if [[ $STATIC_ANALYSIS = 1 ]]; then composer stan; fi
-
-after_success:
-  - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION = 7.3 ]]; then bash <(curl -s https://codecov.io/bash); fi
-
-notifications:
-  email: false
```
dereuromark, line 50: Refs https://github.com/cakephp/authentication/pull/419#pullrequestreview-526926081

othercorey: The email notification?

dereuromark: Wrong line I guess, no, this was referring to the dist XML file to best be kept.

othercorey: I didn't change phpunit.xml.dist.

.gitattributes
```diff
@@ -9,9 +9,8 @@
 .gitignore export-ignore
 .github export-ignore
 .scrutinizer.yml export-ignore
-.travis.yml export-ignore
 appveyor.yml export-ignore
 CONTRIBUTING.md export-ignore
 phpunit.xml.dist export-ignore
-phpcs.xml.dist export-ignore
-tests/test_app/ export-ignore
+phpcs.xml export-ignore
```
dereuromark, line 15: This file
I dont know why my browser jumped to the wrong ones.

.travis.yml
dereuromark: OK, then all good.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,200 @@
...
+          export CODECOVERAGE=1 && vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+        else
+          vendor/bin/phpunit
+        fi
+
+    - name: Submit code coverage
+      if: matrix.php-version == '7.4'
+      uses: codecov/codecov-action@v1
+
+  testsuite-windows:
```
markstory, line 94: Do we need a windows build? I don't think there is anything in this project that is windows specific.

othercorey: If we don't need to run the db queries on windows, then no.

ADmad: We don't need to test db queries on windows. Let's not contribute to global warming unnecessarily.

othercorey: I'm not trying to contribute to global warming here.

othercorey: Removed the windows build

othercorey:
Do we need to test on anything other than sqlite for these?

markstory:
> Do we need to test on anything other than sqlite for these?

I don't think that is necessary right now. We're aren't doing any fancy queries in this plugin.

othercorey:
> > Do we need to test on anything other than sqlite for these?
> 
> I don't think that is necessary right now. We're aren't doing any fancy queries in this plugin.

My point was that we're building on all the dbs, which is why the windows build existed. If we don't need to test all of them this test suite can be smaller.

PR #158, opened 2020-11-22
2.x - Improve Windows compatibility in composer require command.

ndm2:
On the Windows command line the caret is an escape character, thus
composer would not receive the literal character and always install
version 2.0.0.

This will not fix the problem in Powershell, which
would require using additional escape characters that are incompatible
with other environments.

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -9,7 +9,7 @@ Project's ROOT directory (where the **composer.json** file is located)
 
 .. code-block:: shell
 
-    php composer.phar require cakephp/authorization:^2.0
+    php composer.phar require "cakephp/authorization:^2.0"
 
 Load the plugin by adding the following statement in your project's
 ``src/Application.php``::


docs/en/index.rst
```diff
@@ -9,7 +9,7 @@ Project's ROOT directory (where the **composer.json** file is located)
 
 .. code-block:: shell
 
-    php composer.phar require cakephp/authorization:^2.0
+    php composer.phar require "cakephp/authorization:^2.0"
```
dereuromark, line 12: It should be enough to quote the version, no?

ndm2: It would be, AFAIK it's just a matter of preference. I grew accustomed to this variant as the concatenated constraint is what's promoted in the composer docs.

dereuromark: Sure, but I dont know.
This only is relevant if that char is involved, e.g. dev-master and co references are all fine as well.
Adding quoting docs for all cases might be a bit much.

ndm2: I'm not married to it, I'll go with whatever is being preferred, what matters is that people who copy the command get the right version.

markstory: This form is simple, harder to make mistakes with and will work for all environments.

PR #164, opened 2021-04-16
Update middleware.rst

davidspeijer:
More information when implementing both Authentication and Authorization using your User class as the Identity.

### DIFF
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -139,7 +139,7 @@ setup::
 You no longer have to change any existing typehints, and can start using
 authorization policies anywhere you have access to your user.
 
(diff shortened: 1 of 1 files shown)


docs/en/middleware.rst
```diff
@@ -139,6 +139,28 @@ setup::
 You no longer have to change any existing typehints, and can start using
 authorization policies anywhere you have access to your user.
 
+If you also use the Authentication plugin make sure that implement both.::
```
dereuromark, line 142:
```suggestion
If you also use the Authentication plugin make sure to implement both interfaces.::
```

markstory:
Thank you 👍

PR #165, opened 2021-04-22
Rename authorize() to check() and authorizeAction() to checkAction().

othercorey:
This deprecates `authorize()` and `authorizeAction()`.

The documentation needs some more love and there will be more discussion on the API.

### DIFF
--- a/docs/en/component.rst
+++ b/docs/en/component.rst
@@ -45,16 +45,16 @@ the component::
     public function edit($id)
     {
         $article = $this->Articles->get($id);
-        $this->Authorization->authorize($article);
+        $this->Authorization->check($article);
         // Rest of the edit method.
     }
 
-Above we see an article being authorized for the current user. Since we haven't 
+Above we see an article being authorized for the current user. Since we haven't
 specified the action to check the request's ``action`` is used. You can specify
 a policy action with the second parameter::
 
     // Use a policy method that doesn't match the current controller action.
-    $this->Authorization->authorize($article, 'update');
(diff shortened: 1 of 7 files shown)


docs/en/component.rst
```diff
@@ -109,23 +112,23 @@ Example::
...
+        $this->Authorization->check($article);
         ...
     }
 
     public function add()
     {
-        //this will authorize against `insert` model action while being called in `add` controller action.
-        $this->Authorization->authorizeModel(); 
+        // check authorization to access $article with action 'insert'
+        $this->Authorization->check($article);
```
othercorey, line 131: Changed to the code that this seemed to be documenting.

markstory (approved):
Looks good to me.

src/Controller/Component/AuthorizationComponent.php
```diff
@@ -306,7 +342,7 @@ public function authorizeAction(): void
      * @param string $configKey Configuration key with actions.
      * @return bool
      */
-    protected function checkAction(string $action, string $configKey): bool
+    protected function actionConfigured(string $action, string $configKey): bool
```
ADmad, line 345: This method name reads a bit weird but since it's not a public method I don't think we need to bikeshed on it :)

ADmad (approved):

src/Controller/Component/AuthorizationComponent.php
othercorey: It was a quick change so didn't conflict. Hopefully something better comes up.

Issue #170, opened 2021-05-17
Allow passing extra options to applyScope()


othercorey:
When applying a scope to a query, all the available data must be set on the `IdentityInterface` or user object which can be tedious especially if they conflict with fields in the identity/user.

Can we pass an extra parameter of options to `applyScope()`?

```
applyScope(?IdentityInterface $user, string $action, $resource, $options = [])
```

This would function similar to finders that accept an array of options that can be used to build the query.

An example use case is page-specific overrides to user options that are not persisted to the user. These can be manually applied to the query, however, that adds 2 sets of authorization logic to the same query which can introduce bugs.

markstory:
Seems reasonable to me.

PR #171, opened 2021-05-19
Add options to applyScope()

othercorey:
Refs https://github.com/cakephp/authorization/issues/170

Applications that implement `IdentityInterface` will have to add the `array $options = []` parameter.

I don't know if this along with the other api clean up should just go into a 3.0 release or if this is ok for a plugin point release.

### DIFF
--- a/src/IdentityInterface.php
+++ b/src/IdentityInterface.php
@@ -51,9 +51,10 @@ interface IdentityInterface extends ArrayAccess
      *
      * @param string $action The action/operation being performed.
      * @param mixed $resource The resource being operated on.
+     * @param array $options Application-specific scope options
      * @return mixed The modified resource.
      */
-    public function applyScope(string $action, $resource);
+    public function applyScope(string $action, $resource, array $options = []);
 
     /**
      * Get the decorated identity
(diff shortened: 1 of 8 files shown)


ADmad:
> I don't know if this along with the other api clean up should just go into a 3.0 release or if this is ok for a plugin point release.

Those changes a not BC and hence would need a new major release.

othercorey:
> Those changes a not BC and hence would need a new major release.

Agreed. Should we just continue with everything in 2.next for a major release or do I need to split this up?

ADmad:
We can just continue using `2.next`. It's not like we are planning to work on a new minor and new major simultaneously for this plugin.

dereuromark:
But what kind of BC break is this in the end? Looks like an optional param, so even if extended, those would continue to work, no?
One could keep the interface as docblock only to have it less impacting until the major.

In the end, the major isn't that expensive either. I don't think too many plugins or alike have it as a direct dependency (yet).

src/IdentityInterface.php
```diff
@@ -51,9 +51,10 @@ public function canResult(string $action, $resource): ResultInterface;
      *
      * @param string $action The action/operation being performed.
      * @param mixed $resource The resource being operated on.
+     * @param array $options Application-specific scope options
      * @return mixed The modified resource.
      */
-    public function applyScope(string $action, $resource);
+    public function applyScope(string $action, $resource, array $options = []);
```
markstory, line 57: This is the breaking change that I would be concerned about. Rolling a new major is likely our best option here. It is a simple upgrade but one we don't want to surprise people with.

PR #178, opened 2021-07-23
Update Doc for Custom UnauthorizedHandler

LordSimal:
1) I changed the demo config for the already present ``Authorization.Redirect`` Handler because I think redirecting an already logged in user to the login form is not very useful just because the user tried to access something he wasn't allowed to.

2) I adjusted the doc on how to handle unauthorized requests and how to implement a custom handler for that. The slack support channel had numerous people asking how to show flash messages after being redirected by the unauthorized handler.

Sure it could also be implemented in the already present RedirectHandler in the plugin, but any logic beyond that still needs a custom handler and this doc should be more clear for all the people who are not that deeply invested in the whole OOP you need to do here 😉

### DIFF
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -262,7 +262,7 @@ How to create a custom UnauthorizedHandler
 
     class CustomRedirectHandler implements \Authorization\Middleware\UnauthorizedHandler\HandlerInterface
 
-5) Add the flash message logic inside the ``handle()`` method like so::
+#. Add the flash message logic inside the ``handle()`` method like so::
 
     use Authorization\Exception\Exception;
     use Psr\Http\Message\ServerRequestInterface;
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -229,7 +229,7 @@ add other ``Authorization\Exception\Exception`` based exceptions to the
         ForbiddenException::class
     ],
 
-See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
+See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/2.next/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
 
 Configuration options are passed to the handler's ``handle()`` method as the
 last parameter.
@@ -291,5 +291,5 @@ This is, because we extend our handler based on the RedirectHandler present in t
     
 The ``custom_param`` appears in the ``$options`` array given to you in the ``handle()`` function inside your ``CustomRedirectHandler`` if you wish to add some more config parameters to your functionality.
 
-You can look at `CakeRedirectHandler <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php>`__ or `RedirectHandler <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__ 
+You can look at `CakeRedirectHandler <https://github.com/cakephp/authorization/blob/2.next/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php>`__ or `RedirectHandler <https://github.com/cakephp/authorization/blob/2.next/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__ 
 how such a Handler can/should look like.
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -258,7 +258,7 @@ How to create a custom UnauthorizedHandler
 
     class CustomRedirectHandler extends \Authorization\Middleware\UnauthorizedHandler\RedirectHandler
 
-4) If not add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to your class::
+#. If not add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to your class::
 
     class CustomRedirectHandler implements \Authorization\Middleware\UnauthorizedHandler\HandlerInterface
 
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -235,7 +235,7 @@ Configuration options are passed to the handler's ``handle()`` method as the
 last parameter.
 
 Add a flash message after being redirected by an unauthorized request
-------------------------------
+----------------------------------------------------------------------
 
 Currently there is no straightforward way to add a flash message to the unauthorized redirect.
 
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -229,7 +229,7 @@ add other ``Authorization\Exception\Exception`` based exceptions to the
         ForbiddenException::class
     ],
 
-See https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php 
+See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
 
 Configuration options are passed to the handler's ``handle()`` method as the
 last parameter.
--- a/docs/en/middleware.rst
+++ b/docs/en/middleware.rst
@@ -235,9 +235,10 @@ How to create a custom UnauthorizedHandler
 
 1) Create the folder ``src/Middleware/UnauthorizedHandler``
 2) Create a class with the namespace ``Middleware\UnauthorizedHandler`` which ends with the string ``Handler`` (like ``src/Middleware/UnauthorizedHandler/CustomRedirectHandler.php``)
-3) Add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to that class (``implements Authorization\Middleware\UnauthorizedHandler\HandlerInterface``)
+3) If you want to have the basic redirect logic then extend your class with ``extends Authorization\Middleware\UnauthorizedHandler\RedirectHandler``
+4) Add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to your class (``implements Authorization\Middleware\UnauthorizedHandler\HandlerInterface``)
 
-4) Add the flash message logic inside the ``handle()`` method like so::
+5) Add the flash message logic inside the ``handle()`` method like so::
 
     public function handle(Exception $exception, ServerRequestInterface $request, array $options = []): ResponseInterface {
         $response = parent::handle($exception, $request, $options);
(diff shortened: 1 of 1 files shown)


docs/en/middleware.rst
```diff
@@ -201,25 +201,88 @@ Both redirect handlers share the same configuration options:
...
-You can also add your own handler. Handlers should implement
-``Authorization\Middleware\UnauthorizedHandler\HandlerInterface``, be suffixed
-with ``Handler`` suffix and reside under your app's or plugin's
-``Middleware\UnauthorizedHandler`` namespace.
+    
+Handlers catch ONLY those exceptions which extend the 
+``Authorization\Exception\Exception`` class.
+If you want to catch any other exceptions which should
+be handled via your handler they need to be added to
+the ``execeptions`` array
```
ndm2, line 222: That reads a little misleading, as if adding exceptions to the `exceptions` config would change the behavior that only exceptions that do extend `\Authorization\Exception\Exception` are being caught, which isn't the case.

LordSimal: OK, yes sorry.
I meant the default RedirectHandler (and anything extending it) only catch exceptions from that class (or exceptions based on that)
https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php#L61
I will reword that.

ndm2: I probably should have explained my reasoning a little more. The type restriction applies to all handlers, `HandlerInterface::handle()` only accepts that type of exception, also the middleware only catches exceptions of that type in the first place. So no matter the handler config, they can only ever receive exceptions that extend `\Authorization\Exception\Exception`.

LordSimal: Ohhh, so its not a restriction of that specific handler, its a restriction of the plugin. I see

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
-You can also add your own handler. Handlers should implement
-``Authorization\Middleware\UnauthorizedHandler\HandlerInterface``, be suffixed
-with ``Handler`` suffix and reside under your app's or plugin's
-``Middleware\UnauthorizedHandler`` namespace.
+    'exceptions' => [
+        MissingIdentityException::class,
+        ForbiddenException::class
+    ],
+
+See https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php
```
markstory, line 232:
```suggestion
See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
```

We should use a link with text here.

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+Add a flash message after being redirected by an unauthorized request
+------------------------------
+
+Currently there is no straightforward way to add a flash message to the unauthorized redirect.
+
+Therefore you need to create your own Handler which adds th flash message (or any 
+other logic you want to happen on redirect)
+
+How to create a custom UnauthorizedHandler
+------------------------------
```
markstory, line 246:
```suggestion
How to create a custom UnauthorizedHandler
-------------------------------------------
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+
+1) Create the folder ``src/Middleware/UnauthorizedHandler``
+2) Create a class with the namespace ``Middleware\UnauthorizedHandler`` which ends with the string ``Handler`` (like ``src/Middleware/UnauthorizedHandler/CustomRedirectHandler.php``)::
```
markstory, line 249:
```suggestion
#. Create the folder ``src/Middleware/UnauthorizedHandler``
#. Create a class with the namespace ``Middleware\UnauthorizedHandler`` which ends with the string ``Handler`` (like ``src/Middleware/UnauthorizedHandler/CustomRedirectHandler.php``)::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+
+    <?php
+    declare(strict_types=1);
+    namespace App\Middleware\UnauthorizedHandler;
+    
+    class CustomRedirectHandler {...}
+
+3) If you want to have the basic redirect logic then extend your class with::
```
markstory, line 257:
```suggestion
#. If you want to have the basic redirect logic then extend your class with::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+
+    class CustomRedirectHandler extends \Authorization\Middleware\UnauthorizedHandler\RedirectHandler
+
+4) If not add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to your class::
```
markstory, line 261:
```suggestion
#. If not add the ``Authorization\Middleware\UnauthorizedHandler\HandlerInterface`` to your class::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+
+    class CustomRedirectHandler implements \Authorization\Middleware\UnauthorizedHandler\HandlerInterface
+
+5) Add the flash message logic inside the ``handle()`` method like so::
```
markstory, line 265:
```suggestion
#. Add the flash message logic inside the ``handle()`` method like so::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+    use Psr\Http\Message\ServerRequestInterface;
+    use Psr\Http\Message\ResponseInterface;
+
+    public function handle(Exception $exception, ServerRequestInterface $request, array $options = []): ResponseInterface {
+        $response = parent::handle($exception, $request, $options);
+        $request->getFlash()->error('You are not authorized to access that location');
+        return $response;
+    }
+
+6) After you are done with the implementation of that function tell the middleware you want to use your custom handler via::
```
markstory, line 277:
```suggestion
#. After you are done with the implementation of that function tell the middleware you want to use your custom handler via::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,98 @@ Both redirect handlers share the same configuration options:
...
+                ForbiddenException::class
+            ],
+            'custom_param' => true,
+        ],
+    ]));
+    
+The ``custom_param`` appears in the ``$options`` array given to you in the ``handle()`` function inside your ``CustomRedirectHandler``.
+
+You can look at https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php or https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php 
+how such a Handler can/should look like.
```
markstory, line 298:
```suggestion
You can look at `CakeRedirectHandler <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php>`__ or `RedirectHandler <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__ 
how such a Handler can/should look like.
```

docs/en/middleware.rst
markstory: Instead of a 6 step process, you could give the entire class as a single code block and instruct the reader where to save it. That is how examples work elsewhere in the book. Readers often have a hard time figuring out how to assemble a class from small snippets.

LordSimal: Yea, good idea. I will do that

LordSimal: I have now reduced the example to a 2 step process. I hope this is now more clear and easier to understand

docs/en/middleware.rst
LordSimal: I hope my explaination doesn't confuse people more 😅

docs/en/middleware.rst
```diff
@@ -201,25 +201,95 @@ Both redirect handlers share the same configuration options:
...
             'queryParam' => 'redirectUrl',
             'exceptions' => [
                 MissingIdentityException::class,
                 OtherException::class,
             ],
         ],
     ]));
+    
+All handlers get the thrown exception object given as a parameter.
+This excepction will always be an instance of ``Authorization\Exception\Exception``.
```
markstory, line 219:
```suggestion
This exception will always be an instance of ``Authorization\Exception\Exception``.
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,95 @@ Both redirect handlers share the same configuration options:
...
+In this example the ``Authorization.Redirect`` handler just gives you the option to 
+specify which exceptions you want to listen to.
+
+So in this example where we use the ``Authorization.Redirect`` handler we can
+add other ``Authorization\Exception\Exception`` based exceptions to the 
+``execeptions`` array if we wan't to handle them gracefully::
```
markstory, line 225:
```suggestion
``execeptions`` array if we want to handle them gracefully::
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,95 @@ Both redirect handlers share the same configuration options:
...
+
+See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
 
 Configuration options are passed to the handler's ``handle()`` method as the
 last parameter.
 
-Handlers catch only those exceptions which extend the
-``Authorization\Exception\Exception`` class.
+Add a flash message after being redirected by an unauthorized request
+------------------------------
```
markstory, line 238:
```suggestion
Add a flash message after being redirected by an unauthorized request
----------------------------------------------------------------------
```

docs/en/middleware.rst
```diff
@@ -201,25 +201,95 @@ Both redirect handlers share the same configuration options:
...
-You can also add your own handler. Handlers should implement
-``Authorization\Middleware\UnauthorizedHandler\HandlerInterface``, be suffixed
-with ``Handler`` suffix and reside under your app's or plugin's
-``Middleware\UnauthorizedHandler`` namespace.
+    'exceptions' => [
+        MissingIdentityException::class,
+        ForbiddenException::class
+    ],
+
+See the `RedirectHandler source <https://github.com/cakephp/authorization/blob/master/src/Middleware/UnauthorizedHandler/RedirectHandler.php>`__
```
ndm2, line 232: Anyone else think that linking to `master` could cause problems down the road once it will be used for a new major version?

LordSimal: I thought linking to an example in the plugin would be better than copying the source.

ndm2: Sure, my only concern is that the linked code can change to the point of being incompatible, given that it's the master branch. Maybe we should link to tags, or to `x.next` branches? 🤔

Ideally there would be a URL format that would allow linking to the latest tag of a major, but I don't think GitHub supports that.

LordSimal: Sure, I'll do that

markstory:
Thank you 👍

PR #179, opened 2021-07-23
Introduce french translation

zachee54:
I introduce a complete french translation, using the recent spanish docs as model for the configuration in config/all.py and docs.Dockerfile.


markstory:
Thank you 👍

PR #184, opened 2021-09-10
Avoid deprecations under php 8.1

CriztianiX:


othercorey:
This needs the use statement or it doesn’t work.

CriztianiX:
> This needs the use statement or it doesn’t work.

The ci/linter (?) has removed that

https://github.com/cakephp/authorization/pull/184/commits/052a0f4fce311a31adc7106d2a2968aa0c0bbaa8

PR #189, opened 2021-10-29
Start roughing out how github action deploys could work

markstory:
Updating elasticsearch from the container is going to be the hard part as I don't know if the address will be exposed to the build environment.

### DIFF
--- a/docs.Dockerfile
+++ b/docs.Dockerfile
@@ -3,22 +3,41 @@ FROM markstory/cakephp-docs-builder as builder
 
 COPY docs /data/docs
 
+ENV LANGS="en es fr"
+
 # Build docs with sphinx
 RUN cd /data/docs-builder && \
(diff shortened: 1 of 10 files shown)


docs.Dockerfile
```diff
@@ -7,6 +7,10 @@ COPY docs /data/docs
 RUN cd /data/docs-builder && \
   make website LANGS="en es fr" SOURCE=/data/docs DEST=/data/website
 
+# Update elastic search index
+RUN cd /data/docs-builder && \
+  make populate-index LANGS="en es fr" SOURCE=/data/docs ES_HOST="$DOKKU_ELASTICSEARCH_AQUA_URL" SEARCH_URL_PREFIX="/authorization/2"
```
markstory: I don't think `DOKKU_ELASTICSEARCH_AQUA_URL` will be set, nor do I yet know how I will set the appropriate value either as a dokku build option or some other sorcery.

PR #192, opened 2022-02-14
move unauthorizedHandler to AuthorizationService

saeideng:


src/AuthorizationService.php
```diff
@@ -22,9 +22,21 @@
...
+use Cake\Core\InstanceConfigTrait;
 
 class AuthorizationService implements AuthorizationServiceInterface
 {
+    use InstanceConfigTrait;
+
+    /**
+     * Default config.
+     *
+     * @var array
```
dereuromark, line 34:
```suggestion
     * @var array<string, mixed>
```

src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php
```diff
@@ -25,6 +25,7 @@
  * This handler will redirect the response if one of configured exception classes is encountered.
  *
  * CakePHP Router compatible array URL syntax is supported.
+ * @deprecated use RedirectHandler instead.
```
markstory, line 28: :clap: I agree with the service hook method developers can use `Router::url()` to generate the URL they need.

saeideng:
do we need move all configs from  `AuthorizationMiddleware` to `AuthorizationService`
or only the `unauthorizedHandler` ?

my opinion : moving all

markstory:
> my opinion : moving all

Would you also be moving all the logic related to these options to the service as well?

saeideng:
> > my opinion : moving all
> 
> Would you also be moving all the logic related to these options to the service as well?


not sure , but I can try it

ADmad:
I didn't see this PR earlier. #211 adds the `unauthorizedHandler` to both the middlewares using a trait instead.

PR #193, opened 2022-03-03
Update policies.rst

passchn:
The Policy class name must contain "Table" - it does not work otherwise. 

It is resolved in src/Policy/OrmResolver.php, getRepositoryPolicy() and $name contains "Table".


passchn:
I'm on 1.x branch because I'm working on an older CakePHP 3.x project.

PR #197, opened 2022-03-26
update to PHPStan 1.5 and Psalm to 4.22

LordSimal:
Here I have a little bit of a problem because I don't know how to resolve those 2 pslam errors:

```
ERROR: NonInvariantDocblockPropertyType - src/Identity.php:32:15 - Property Authorization\Identity::$identity has type Authentication\IdentityInterface, not invariant with Authorization\IdentityDecorator::$identity of type ArrayAccess|array<array-key, mixed> (see https://psalm.dev/267)
    protected $identity;


ERROR: DocblockTypeContradiction - src/Policy/OrmResolver.php:84:63 - Docblock-defined type Cake\Datasource\QueryInterface for $resource is always object (see https://psalm.dev/155)
        $name = is_object($resource) ? get_class($resource) : gettype($resource);
```


dereuromark:
> public function getPolicy($resource)

We probably have to wait for being able to add `object` type here, then the additional internal check could be removed.
Technically I guess one would need to add `mixed` to the docblock, but I am not sure we should.
Maybe it is better to just silence this one.

ADmad:
> ERROR: NonInvariantDocblockPropertyType - src/Identity.php:32:15 - Property Authorization\Identity::$identity has type Authentication\IdentityInterface, not invariant with Authorization\IdentityDecorator::$identity of type ArrayAccess|array<array-key, mixed> (see https://psalm.dev/267)
    protected $identity;

This error is valid since in the parent class `IdentityDecorator::$identity` has type `array|\ArrayAccess` while `Identity::$identity` has type `IdentityInterface`. Not sure how to resolve this without using another property name in `Identity`.

> ERROR: DocblockTypeContradiction - src/Policy/OrmResolver.php:84:63 - Docblock-defined type Cake\Datasource\QueryInterface for $resource is always object (see https://psalm.dev/155)
        $name = is_object($resource) ? get_class($resource) : gettype($resource);

The type checks won't be required when eventually argument types are added.

LordSimal:
I rebased this branch and updated to latest PHPStan and psalm but of course we still have the same main problem as discussed above

LordSimal:
seems like I did something majorly wrong... let me redo that

PR #199, opened 2022-04-18
Translated index and policy.

yuki-yasumura:
The others will be submitted in English once the ja translation was not there in the first place.
By being present in English, someone can quickly fork and correct it. (I will be translating it myself. I can't do it right now because I can't get around to it.)

### DIFF
--- a/docs/ja/2-0-migration-guide.rst
+++ b/docs/ja/2-0-migration-guide.rst
@@ -1,0 +1,18 @@
+2.0 Migration Guide
+###################
+
+Authorization 2.0 contains new features and a few breaking changes.
+
+Breaking Changes
+================
+
+The ``IdentityInterface`` has had typehinting added. If you have implemented the
+``IdentityInterface`` you will need to update your application's implementation
+to reflect the new typehints.
+
+In addition to typehints ``IdentityInterface`` has a ``canResult()`` method
+added. This method always returns a ``ResultInterface`` object while ``can()``
+always returns a boolean. In 1.x the ``can()`` method would return a boolean or
+``ResultInterface`` depending on what the policy returned. This made knowing the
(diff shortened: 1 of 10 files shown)


docs/ja/2-0-migration-guide.rst
```diff
@@ -0,0 +1,18 @@
+2.0 Migration Guide
```
markstory, line 1: Could we get this english content removed? We generally only keep localized content in the `ja` directories. Providing users with potentially incorrect documentation is far worse than having users switch the language code in the URL or use a different tab with the english docs is.

If that proves to be challenging because of linking, I can take care of finding a solution to the incomplete translation that won't get outdated on my side. I have a half baked idea that involves using `make` to copy all the english content over if files do not exist :smiling_imp:

docs/ja/contents.rst
```diff
@@ -0,0 +1,15 @@
...
+    :caption: CakePHP Authorization
+
+    /index
+    /policies
+    /policy-resolvers
+    /middleware
+    /checking-authorization
+    /component
+    /request-authorization-middleware
+    /2-0-migration-guide
```
markstory, line 15:
```suggestion
```
To go with the suggestion of removing the un-translated content.

yuki-yasumura:
@markstory 

I see...

Yes, sir. Then I shall translate it all! It may take me until next week or so, but please bear with me :bow:

markstory:
Take your time. I'm planning on spending some time later this week to figure out a better solution for partial translations. The process we've been using isn't great as it requires either removing links or adding english pages to translated content. I'll post an update if I figure out a better solution.

yuki-yasumura:
@markstory 

Sorry, I am late.
All translated.

よろしくお願いします！！ 🍣

markstory:
Thank you 🎉

PR #208, opened 2022-06-29
Cake5 updates

ADmad:


src/Middleware/AuthorizationMiddleware.php
```diff
@@ -78,23 +78,10 @@ class AuthorizationMiddleware implements MiddlewareInterface
...
-            !$subject instanceof AuthorizationServiceProviderInterface
-        ) {
-            $expected = implode('` or `', [
-                AuthorizationServiceInterface::class,
-                AuthorizationServiceProviderInterface::class,
-            ]);
-            $type = is_object($subject) ? get_class($subject) : gettype($subject);
-            $message = sprintf('Subject must be an instance of `%s`, `%s` given.', $expected, $type);
-
-            throw new InvalidArgumentException($message);
```
markstory, line 95: :clap: So nice that the type system can take care of this now.

markstory (approved):

composer.json
```diff
@@ -77,5 +78,7 @@
             "cakephp/plugin-installer": true,
             "dealerdirect/phpcodesniffer-composer-installer": true
         }
-    }
+    },
+    "minimum-stability": "dev",
```
othercorey, line 82: Is this needed with explicit versions?

ADmad: Yes it's needed to be able to install indirect non-stable dependencies. In this particular case `bake` has `twig-view` with `dev-cake5`.

othercorey: Ah right.

PR #209, opened 2022-08-01
unauthorizedHandler config for RequestAuthorizationMiddleware

### LINKED ISSUE #120: RequestAuthorizationMiddleware config redirect url missing
I try to use to add the RequestAuthorizationMiddleware for some controller in my application in cake 4.
In the [Authorization Middleware](https://book.cakephp.org/authorization/2/en/middleware.html) i see a code block like this:

```php
$middlewareQueue->add(new AuthorizationMiddleware($this, [
    'unauthorizedHandler' => [
        'className' => 'Authorization.Redirect',
        'url' => '/users/login',
        'queryParam' => 'redirectUrl',
        'exceptions' => [
            MissingIdentityException::class,
            OtherException::class,
        ],
    ],
]));
```
I can add a redirect url if the authorization failed. 
This work fine but how i can add a redirect url for the [Request Authorization Middleware](https://book.cakephp.org/authorization/2/en/request-authorization-middleware.html) ?

matteorebeschi:
This should fix #120

### DIFF
--- a/src/Middleware/AuthorizationMiddleware.php
+++ b/src/Middleware/AuthorizationMiddleware.php
@@ -1,5 +1,4 @@
 <?php
-
 declare(strict_types=1);
 
 /**
--- a/src/Middleware/AuthorizationMiddleware.php
+++ b/src/Middleware/AuthorizationMiddleware.php
@@ -38,6 +38,29 @@ use RuntimeException;
 class AuthorizationMiddleware extends BaseAuthorizationMiddleware
 {
     /**
+     * Default config.
+     *
+     * - `identityDecorator` Identity decorator class name or a callable.
+     *   Defaults to IdentityDecorator
+     * - `identityAttribute` Attribute name the identity is stored under.
+     *   Defaults to 'identity'
+     * - `requireAuthorizationCheck` When true the middleware will raise an exception
+     *   if no authorization checks were done. This aids in ensuring that all actions
+     *   check authorization. It is intended as a development aid and not to be relied upon
+     *   in production. Defaults to `true`.
+     * - `unauthorizedHandler` Allows to define a custom behavior when the authorization check fails
+     *   Defaults to 'Authorization.Exception'
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'identityDecorator' => null,
+        'identityAttribute' => 'identity',
+        'requireAuthorizationCheck' => true,
+        'unauthorizedHandler' => 'Authorization.Exception',
+    ];
+
+    /**
      * Authorization service or application instance.
      *
      * @var \Authorization\AuthorizationServiceInterface|\Authorization\AuthorizationServiceProviderInterface
--- a/src/Middleware/AuthorizationMiddleware.php
+++ b/src/Middleware/AuthorizationMiddleware.php
@@ -1,4 +1,5 @@
 <?php
+
 declare(strict_types=1);
 
 /**
@@ -14,6 +15,7 @@ declare(strict_types=1);
  * @since         1.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Authorization\Middleware;
 
 use Authentication\IdentityInterface as AuthenIdentityInterface;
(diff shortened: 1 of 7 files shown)


src/Middleware/AuthorizationMiddleware.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
markstory: Changing the formatting is going to cause CI to fail. Could you use `vendor/bin/phpcbf` to reformat the modified files to match the project's coding conventions?

src/Middleware/AuthorizationMiddleware.php
```diff
@@ -143,43 +116,22 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+    protected function getAuthorizationService(ServerRequestInterface $request): AuthorizationServiceInterface
+    {
         $service = $this->subject;
         if ($this->subject instanceof AuthorizationServiceProviderInterface) {
             $service = $this->subject->getAuthorizationService($request);
         }
 
         if (!$service instanceof AuthorizationServiceInterface) {
-            throw new RuntimeException(sprintf(
+            throw new \RuntimeException(sprintf(
```
markstory: We prefer to have `use RuntimeException` at the top of the file instead of inline qualifiers.

src/Middleware/BaseAuthorizationMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'identityDecorator' => null,
+        'identityAttribute' => 'identity',
+        'requireAuthorizationCheck' => true,
+        'unauthorizedHandler' => 'Authorization.Exception',
+        'authorizationAttribute' => 'authorization',
+        'method' => 'access',
```
markstory: This configuration option isn't used by all subclasses, could it be defined in only the class that uses it?

src/Middleware/BaseAuthorizationMiddleware.php
```diff
@@ -0,0 +1,100 @@
...
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use RuntimeException;
+
+/**
+ * Authorization Middleware.
+ *
+ * Injects the authorization service and decorated identity objects into the request object as attributes.
+ */
+abstract class BaseAuthorizationMiddleware implements MiddlewareInterface
```
markstory, line 34: I'm not sure we *need* a base class. Having two copies of the `getHandler()` method is a better solution in my opinion as it uses fewer lines of code, and doesn't introduce options that aren't shared.

tests/test_app/TestApp/Middleware/UnauthorizedHandler/SuppressHandler.php
```diff
@@ -14,13 +15,14 @@
...
  */
+
 namespace TestApp\Middleware\UnauthorizedHandler;
 
 use Authorization\Exception\Exception;
 use Authorization\Middleware\UnauthorizedHandler\HandlerInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
-use Zend\Diactoros\Response;
+use Cake\Http\Response;
```
markstory, line 25: Why did this change?

matteorebeschi: Because PHPUnit tests were failing because of that missing class. The Zend\Diactoros package is not included in the composer.json.

src/Middleware/BaseAuthorizationMiddleware.php
matteorebeschi: I am generally under the impression that having a base class to share the common code is better than having the same exact method(s) copypasted over and over, but if you want I can remove it.

src/Middleware/AuthorizationMiddleware.php
matteorebeschi: Done.

src/Middleware/AuthorizationMiddleware.php
matteorebeschi: Done.

tests/test_app/TestApp/Middleware/UnauthorizedHandler/SuppressHandler.php
ADmad: @markstory Laminas removed the Zend => Laminas compatibility classes in a recent release.

src/Middleware/BaseAuthorizationMiddleware.php
matteorebeschi: I've redefined the $_defaultConfig array on each class based on what each class specifically needs as it was before.

ADmad:
Closed by #211

PR #212, opened 2022-10-08
cake5 updates

LordSimal:


psalm.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <psalm
-    totallyTyped="false"
+    errorLevel="2"
```
othercorey: why level 2?

LordSimal: Because its the replacement for `totallyTyped="false"` according to: https://psalm.dev/docs/running_psalm/configuration/#totallytyped

`errorLevel="1"` reported even more errors.

othercorey: authentication doesn't need this, what is the default? I'd prefer to have these common settings the same everywhere.

ADmad: Level 2 is too low, for e.g. the core uses level 4.

LordSimal: The default `errorLevel` is 2. We can also drop the `resolveFromConfigFile` definition because we also use the default value.
image

LordSimal: @ADmad Psalm error levels are stricter when going lower. See https://psalm.dev/docs/running_psalm/error_levels/
So level 1 is stricter than level 2.

ADmad: Sorry again :)  I was confused as phpstan's level become stricter with higher number.

PR #215, opened 2022-11-09
Fix ControllerHookPolicy Code Example

amayer5125:
The old example had incorrect parameters for the [__call](https://www.php.net/manual/en/language.oop5.overloading.php#object.call) method. The first param is the name of the method being called. The second param is all the arguments that were passed to the called method.

We are also given an identity, not an array which is what `isAuthorized` expects. To fix this we can pass `$user->getOriginalData()` to the old `isAuthorized` methods.


markstory:
Thank you 🎉

PR #217, opened 2022-12-21
Developer friendly exception message

asgraf:
Currently RequestAuthorizationMiddleware throws an exception with not very telling message:
```
Error
```
After this little patch you will get more descriptive exception message similar to:
```
Identity is not authorized to perform access on /admin/.
```


markstory:
Thank you 🎉

PR #218, opened 2022-12-21
3.x: fix basepath in redirecthandler

LordSimal:
The current tests for 3.x fail due to the fact that the basepath is not handled correctly.
```
1) Authorization\Test\TestCase\Middleware\UnauthorizedHandler\CakeRedirectHandlerTest::testHandleRedirectWithBasePath
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'/basedir/login?redirect=%2Fadmin%2Fdashboard'
+'/login?redirect=%2Fadmin%2Fdashboard'
```

### DIFF
--- a/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php
+++ b/src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php
@@ -77,6 +77,11 @@ class CakeRedirectHandler extends RedirectHandler
             $url['?'][$options['queryParam']] = $redirect;
         }
 
+        $requestBase = $request->getAttribute('base');
+        if ($requestBase) {
+            $url['_base'] = $requestBase;
+        }
+
         return Router::url($url);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Middleware/UnauthorizedHandler/CakeRedirectHandler.php
```diff
@@ -77,6 +77,11 @@ protected function getUrl(ServerRequestInterface $request, array $options): stri
             $url['?'][$options['queryParam']] = $redirect;
         }
 
+        $requestBase = $request->getAttribute('base');
+        if ($requestBase) {
+            $url['_base'] = $requestBase;
+        }
+
         return Router::url($url);
```
ndm2, line 85: Shouldn't `Router::url()` here handle the base path? The test might just need to use `Router::setRequest()`?

ADmad:
I did a 2.x => 3.x merge and fixed this with other issues. Using `Router::setRequest()` is the proper way to fix it.

Issue #225, opened 2023-01-05
Allow multiple optional parameters to be passed down to applyScope()


LordSimal:
To adjust ORM queries one has to use the `->applyScope()` method of the [AuthorizationComponent](https://book.cakephp.org/authorization/2/en/component.html#applying-policy-scopes).

But in certain scenarios it would be helpfull to pass down arguments to the `->applyScope()` method because the query needs to be adjusted according to
- the current logged in user AND
- the state of a specific service object or maybe request parameters.

The proposed solution would look like that: 
```
public function index(MyService $myService)
{
    $this->Authorization->skipAuthorization();
    $query = $this->Products->find();
    $query = $this->Authorization->applyScope($query, 'index', $myService);
    $products = $this->paginate($query);

    $this->set(compact('products'));
}
```

And in the `ProductsTablePolicy.php` it would look like this:
```
class ProductsTablePolicy
{
    public function scopeIndex(IdentityInterface $user, Query $query, $myService)
    {
        return $query->where(['Products.price >' => 0]);
    }
}
```

This would be of course a BC break because we would have to adjust at least the `IdentityInterface::applyScope` as well as the `AuthorizationServiceInterface::applyScope`  

What do you think?

PR #226, opened 2023-01-06
3.x: add ability to add multiple optional parameters to table scopes

LordSimal:
Fixes https://github.com/cakephp/authorization/issues/225

### DIFF
--- a/src/IdentityDecorator.php
+++ b/src/IdentityDecorator.php
@@ -76,9 +76,9 @@ class IdentityDecorator implements IdentityInterface
     /**
      * @inheritDoc
      */
-    public function applyScope(string $action, $resource): mixed
+    public function applyScope(string $action, $resource, mixed ...$optionalArgs): mixed
     {
-        return $this->authorization->applyScope($this, $action, $resource);
+        return $this->authorization->applyScope($this, $action, $resource, ...$optionalArgs);
     }
 
     /**
--- a/src/IdentityDecorator.php
+++ b/src/IdentityDecorator.php
@@ -76,7 +76,7 @@ class IdentityDecorator implements IdentityInterface
     /**
      * @inheritDoc
      */
-    public function applyScope(string $action, $resource, mixed ...$optionalArgs): mixed
+    public function applyScope(string $action, mixed $resource, mixed ...$optionalArgs): mixed
     {
         return $this->authorization->applyScope($this, $action, $resource, ...$optionalArgs);
     }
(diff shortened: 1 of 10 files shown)


src/IdentityDecorator.php
```diff
@@ -76,9 +76,9 @@ public function canResult(string $action, $resource): ResultInterface
     /**
      * @inheritDoc
      */
-    public function applyScope(string $action, $resource): mixed
+    public function applyScope(string $action, $resource, mixed ...$optionalArgs): mixed
```
ADmad:
```suggestion
    public function applyScope(string $action, mixed $resource, mixed ...$optionalArgs): mixed
```
Seems the type for `$resource` was missed earlier.

src/IdentityInterface.php
```diff
@@ -53,9 +53,10 @@ public function canResult(string $action, mixed $resource): ResultInterface;
      *
      * @param string $action The action/operation being performed.
      * @param mixed $resource The resource being operated on.
+     * @param mixed $optionalArgs Multiple additional arguments which are passed to the scope
      * @return mixed The modified resource.
      */
-    public function applyScope(string $action, mixed $resource): mixed;
+    public function applyScope(string $action, mixed $resource, mixed ...$optionalArgs): mixed;
```
markstory, line 59: This will be a breaking change as user-land code will have implemented this interface.  We should add this to the documentation/migration guide.

LordSimal: Indeed. I was just curious if you guys even like this idea or not 😄 
The first version of the 3.0 migration guide has been added with the major changes I have seen via looking at https://github.com/cakephp/authorization/compare/2.x...3.x-scope-params

tests/TestCase/AuthorizationServiceTest.php
```diff
@@ -173,6 +176,33 @@ public function testApplyScopeMethodMissing()
...
+
+        $query->expects($this->once())
+            ->method('where')
+            ->with([
+                'identity_id' => 9,
+                'firstArg' => 'first argument',
+                'secondArg' => false,
+            ])
+            ->willReturn($query);
+        $result = $service->applyScope($user, 'additionalArguments', $query, 'first argument', false);
```
markstory, line 201: Will these changes also enable the use of named parameters? eg. `applyScope($user, category: $category)`

LordSimal: I have added a test which proves it works 😉

markstory: Yay!

PR #227, opened 2023-01-09
Correct PHP requirement

ravage84:
According to https://github.com/cakephp/authorization/releases/tag/2.3.0

> Authorization 2.3.0 requires at least CakePHP 4.4.0 and PHP 7.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     ],
     "type": "cakephp-plugin",
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.4",
         "cakephp/http": "^4.0",
         "psr/http-client": "^1.0",
         "psr/http-message": "^1.0",


ADmad:
The test suite ran fine on PHP 7.2 https://github.com/cakephp/authorization/actions/runs/3789799126/jobs/6443887879.

So not sure if the release description is accurate.

othercorey:
The cake versions weren't updated either so something's amiss.

markstory:
Could be that I conflated authentication & authorization and thought that both of them had dependency updates. I think our best course of action is to update the release notes instead of changing the code. Unless we want to start pruning old PHP versions as we build towards 5.x

othercorey:
Let's do that. I'd rather not restrict plugins until needed.

ADmad:
I have updated the release notes.

PR #229, opened 2023-01-10
Declare some missing dynamic properties in tests

jamisonbryant:
Adds the required trait to suppress deprecation warnings related to dynamic props in the plugin test suite.

I figured the trait was the best solution since these are just test cases, however if it would be better for me to just declare the props explicitly (& well-typed) I'm happy to do that instead.


ADmad:
It would be better to declare the properties used instead.

LordSimal:
Make sure to run `composer cs-fix` to auto-fix CS problems.
But in the end `composer cs-check` should all be green before you commit.

markstory:
Thank you 🎉

PR #231, opened 2023-03-03
Update policies.rst

geoidesic:
Added a link to applying policies as it's not obvious from this page that such is necessary, nor how to do so.

### DIFF
--- a/docs/en/policies.rst
+++ b/docs/en/policies.rst
@@ -134,4 +134,4 @@ Before hooks are expected to return one of three values:
   
 Applying Policies
 -----------------
-https://book.cakephp.org/authorization/2/en/component.html#applying-policy-scopes
+See :ref:`applying-policy-scopes` for how to apply policies in your controller actions.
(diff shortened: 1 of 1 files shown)


docs/en/policies.rst
```diff
@@ -131,3 +131,7 @@ Before hooks are expected to return one of three values:
 - ``false`` The user is not allowed to proceed with the action.
 - ``null`` The before hook did not make a decision, and the authorization method
   will be invoked.
+  
+Applying Policies
+-----------------
+https://book.cakephp.org/authorization/2/en/component.html#applying-policy-scopes
```
markstory:
```suggestion
See :ref:`applying-policy-scopes` for how to apply policies in your controller actions.
```

We should be able to use a `ref` directive to create a link.

ndm2: And shouldn't this rather be a note in the "Policy Scopes" section, given that this specifically links to applying scopes?

btw, everything after "Policy Result Objects" seems to be incorrectly nested.

markstory: > btw, everything after "Policy Result Objects" seems to be incorrectly nested.

I can get that fixed outside of these changes.

PR #236, opened 2023-05-01
3.x: add ability to add multiple optional parameters to can and canResult

### LINKED ISSUE #135: Authorization Identity can() method need to allow second parms to optional.
I am using Authorization plugin and created Policies and these policies working good in the controller like can index/add/edit/delete but i have issue in the view index.php there i need to hide add button using `can()` method like below
```

if($this->request->getAttribute('identity')->can('add')){
         echo $this->Html->link("<i class=\"fa fa-plus\"></i> " . __('New User'), ["action" => "add"], ["escape" => false]);
  }
```

There we need second parameter `newEmptyEntity` object but i am in index function so i don't want to add `newEmptyEntity` object. when i used above script for hide add button then this error will be appear 

> (Error: Too few arguments to function Authorization\IdentityDecorator::can(), 1 passed in  index.php on line 21 and exactly 2 expected.)

 My policy method is this one:

```
public function canAdd(IdentityInterface $user)
    {
        if($user->is_supper_admin){
            return true;
        }
        return false;
    }
```

In this case i don't need second parameter so please resolve this issue

image

Thanks

LordSimal:
Resolves #135 
Refs https://github.com/cakephp/authorization/pull/226

### DIFF
--- a/docs/en/3-0-migration-guide.rst
+++ b/docs/en/3-0-migration-guide.rst
@@ -14,11 +14,14 @@ The following interfaces now have appropriate parameter and return types added:
 - ``RequestPolicyInterface.php``
 - ``ResolverInterface``
 
-Multiple optional arguments for ``applyScope``
-----------------------------------------------
+Multiple optional arguments for ``applyScope``, ``can`` and ``canResult``
+-------------------------------------------------------------------------
 
-``IdentityInterface::applyScope`` as well as ``AuthorizationServiceInterface::applyScope``
-allow multiple optional arguments to be added.
+The following interface methods have been adjusted to pass on multiple optional arguments.
+- ``IdentityInterface::applyScope``
+- ``AuthorizationServiceInterface::applyScope``
+- ``AuthorizationServiceInterface::can``
+- ``AuthorizationServiceInterface::canResult``
 
 Removed methods
 ---------------
--- a/tests/TestCase/AuthorizationServiceTest.php
+++ b/tests/TestCase/AuthorizationServiceTest.php
@@ -66,6 +66,50 @@ class AuthorizationServiceTest extends TestCase
         $this->assertTrue($result);
     }
 
+    public function testCanWithAdditionalParams()
+    {
+        $resolver = new MapResolver([
+            Article::class => ArticlePolicy::class,
+        ]);
+
+        $service = new AuthorizationService($resolver);
+
+        $user = new IdentityDecorator($service, [
+            'role' => 'admin',
+        ]);
+
+        $innerService = function () {
+            return true;
+        };
+
+        $result = $service->can($user, 'withService', new Article(), $innerService);
+        $this->assertTrue($result);
+    }
+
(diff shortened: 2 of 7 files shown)


docs/en/3-0-migration-guide.rst
```diff
@@ -14,11 +14,14 @@ The following interfaces now have appropriate parameter and return types added:
 - ``RequestPolicyInterface.php``
 - ``ResolverInterface``
 
-Multiple optional arguments for ``applyScope``
+Multiple optional arguments for ``applyScope``, ``can`` and ``canResult``
 ----------------------------------------------
```
ADmad: I believe the underline length needs to match the text length above.

LordSimal: The amount of times I have falen into this RST trap is boggling me.... thanks 😅

tests/TestCase/AuthorizationServiceTest.php
```diff
@@ -66,6 +66,50 @@ public function testCan()
...
+
+        $innerService1 = function () {
+            return true;
+        };
+
+        $innerService2 = function () {
+            return false;
+        };
+
+        $result = $service->can(user: $user, action: 'witMultipleServices', resource: new Article(), service2: $innerService2, service1: $innerService1);
```
markstory:
```suggestion
        $result = $service->can(user: $user, action: 'withMultipleServices', resource: new Article(), service2: $innerService2, service1: $innerService1);
```

tests/TestCase/AuthorizationServiceTest.php
```diff
@@ -82,6 +126,50 @@ public function testCanWithResult()
...
+
+        $innerService1 = function () {
+            return true;
+        };
+
+        $innerService2 = function () {
+            return false;
+        };
+
+        $result = $service->canResult(user: $user, action: 'witMultipleServices', resource: new Article(), service2: $innerService2, service1: $innerService1);
```
markstory:
```suggestion
        $result = $service->canResult(user: $user, action: 'withMultipleServices', resource: new Article(), service2: $innerService2, service1: $innerService1);
```

tests/test_app/TestApp/Policy/ArticlePolicy.php
```diff
@@ -116,4 +117,14 @@ public function canPublish($user, Article $article)
 
         return new Result($article->get('user_id') === $user['id']);
     }
+
+    public function canWithService($user, Article $article, Closure $service)
+    {
+        return $service();
+    }
+
+    public function canWitMultipleServices($user, Article $article, Closure $service1, Closure $service2)
```
markstory:
```suggestion
    public function canWithMultipleServices($user, Article $article, Closure $service1, Closure $service2)
```

tests/TestCase/AuthorizationServiceTest.php
LordSimal: My dyslexia has struck again... thanks 😁

PR #245, opened 2023-08-21
Add support for applyScope with missing identity in AuthorizationComponent

robertpustulka:


othercorey:
This feels like a minor release not a bug fix release.

PR #256, opened 2023-10-10
Fix Applying Policy Scopes Link

amayer5125:
The link wasn't working before. I tried to get the reference working, but I couldn't figure it out so I replaced it with a link.


markstory:
Thank you 🎉

PR #257, opened 2023-10-13
Set missing reference for applying-policy-scopes

ravage84:
The reference was not converted into a link in https://book.cakephp.org/authorization/2/en/policies.html#applying-policies

grafik


ndm2:
~You're a little too late https://github.com/cakephp/authorization/commit/ee1dd1564d550597ac6f39842e25d333b6171a41~

nvm, this is 2.x not 3.x 🤦

Issue #259, opened 2023-11-01
Feature request: Allow Table policies to use BeforePolicyInterface 


jamisonbryant:
Hello, I noticed while implementing a Table policy recently (specifically a [Policy Scope](https://book.cakephp.org/authorization/2/en/policies.html#policy-scopes)) I can't attach the `BeforePolicyInterface` and expect the `before()` method to get called prior to the execution of the `scopeFoo()` method.

It appears as if this is because the `AuthorizationService::applyScope()` method does not check if the policy is an instance of `BeforePolicyInterface` like `performCheck()` does. But other than that, the methods look pretty darn similar.

Was this decision to not check for the Interface by design? If so, what was the motivation? And if not, would this change be beneficial? I could certainly use it to allow my superadmins to skip scope limitations on Queries that I apply to normal users.

Also happy to take a first stab at a PR, if this change is OK with the maintainers.

markstory:
> Was this decision to not check for the Interface by design? If so, what was the motivation? And if not, would this change be beneficial? 

No it wasn't intentional, using the same interface before `applyScope` could be ok. If a policy implements both record policy methods and `applyScope` though the `before` method would become messy. Should we have a new method for `beforeScope`?

Adding methods to existing interfaces isn't possible because of backwards compatibility, but we could add a new interface for `BeforeScopePolicy` and merge the two interfaces in the future :shrug:

jamisonbryant:
> we could add a new interface for `BeforeScopePolicy`

This I like. I'm happy to tackle this, if desired. I don't think it would be super difficult.

> and merge the two interfaces in the future 🤷

Would this even be strictly necessary? I would think it would be OK to acknowledge that although a scope _is_ a policy, it's a different _kind_ of policy and so a different interface is necessary.

jamisonbryant:
Description for the `beforeScope` method (first attempt):

>  If boolean value is returned, the scope application will be skipped and the unmodified resource will be returned. In case of `null`, the scope will be applied.

I have two questions:

1. I'm getting hung up on "boolean value" because technically a `ResultInterface` is not a boolean. Unless it is understood that a `Result` with status = true is the same as boolean true, and same for status = false?
2. Assuming the answer to the previous question is 'yes', what should be done when true or false are returned? It makes sense that returning `null` means the scope application will still be run. But in the context of a scope, I'm not sure what it should do when a boolean is returned.

markstory:
[...]

PR #260, opened 2023-11-08
Feature: BeforeScopeInterface (3.next)

jamisonbryant:
## Description

In our app we would like the ability to run pre-authorization checks on Scopes, something that is not currently possible because the `BeforePolicyInterface` doesn't work on Scopes. This PR adds the `BeforeScopeInterface` and updates the `AuthorizationService` to check for it and call `beforeScope()` just like it does for BeforePolicyInterface and `before()`.

References #259 

## Changes Made

- Added the `BeforeScopeInterface` with a single method `beforeScope()`.
- Updated the `AuthorizationService::applyScope()` method to check for `BeforeScopeInterface` and call `beforeScope()` prior to passing control to the scope handler.
- Updated tests.

## Benefits

- Adds the ability to run logic before scopes are applied, something that is not currently possible with the plugin.

## Checklist

- [x] Tests written and passing
- [x] Code follows coding standards
- [x] Documentation updated if necessary - NOW DONE
- [x] Spelling/grammar check done

### DIFF
--- a/docs/en/policies.rst
+++ b/docs/en/policies.rst
@@ -132,6 +132,31 @@ Before hooks are expected to return one of three values:
 - ``null`` The before hook did not make a decision, and the authorization method
   will be invoked.
 
+Scope Pre-conditions
+====================
+
+Like policies, scopes can also define pre-conditions. These are useful when you
+want to apply common conditions to all scopes in a policy. To use pre-conditions
+on scopes you need to implement the ``BeforeScopeInterface`` in your scope policy::
+
+    namespace App\Policy;
+
+    use Authorization\Policy\BeforeScopeInterface;
+
+    class ArticlesTablePolicy implements BeforeScopeInterface
+    {
+        public function beforeScope($user, $query, $action)
+        {
+            if ($user->getOriginalData()->is_trial_user) {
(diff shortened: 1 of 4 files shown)


jamisonbryant:
_Dreadfully_ sorry for being rude, but is there anything I can do to expedite review & merge of this branch and #261 for this feature? (already rebased it)

My job needs this feature, and I am leaving for paternity leave after Friday. I would love to have this merged in (even to a n.next branch) so they do not have to use my fork for the next couple months :)

LordSimal:
I would like to leave this up to @ADmad and/or @markstory how they want to handle this, but usually features which were added in previous versions will be merged "upwards" so if we get the 2.next PR in then this 3.next PR is not needed.

So basically if you get your PR fixed up to not have unnencessary commits in them, then we can merge it to 2.next and this feature will be added to the next minor release of this plugin (2.x as well as 3.x)

But in the meantime I would say its totally fine to keep your forked branch in the composer.json so that at least your feature works. Your co-workers will just have to adjust the composer.json after this feature has been released.

markstory:
With #264 merged, I'm going to close this as the 2->3 merge will take care of getting this to 3.x

PR #262, opened 2023-11-10
3.x: use union types

Indianer3c:


docs/en/request-authorization-middleware.rst
```diff
@@ -33,7 +33,7 @@ and add::
          * @param \Cake\Http\ServerRequest $request Server Request
          * @return bool
          */
-        public function canAccess($identity, ServerRequest $request)
+        public function canAccess($identity, ServerRequest $request): ResultInterface
```
dereuromark: Why does a bool method (and documented as such) now return an object?

LordSimal: this doc just needs to be updated to represent https://github.com/cakephp/authorization/blob/3.x/src/Policy/RequestPolicyInterface.php#L34

Indianer3c: Sorry my mistake

LordSimal:
thank you 👍🏻

PR #272, opened 2024-01-05
3.x: adjust docs for user entity implements IdentityInterface

LordSimal:


docs/en/middleware.rst
```diff
@@ -80,33 +80,33 @@ implementing the ``Authorization\IdentityInterface`` and using the
     class User extends Entity implements IdentityInterface
     {
         /**
-         * Authorization\IdentityInterface method
+         * @inheritDoc
          */
-        public function can($action, $resource): bool
+        public function can(string $action, $resource): bool
```
dereuromark:
```suggestion
        public function can(string $action, mixed $resource): bool
```
According to https://github.com/cakephp/authorization/blob/3.x/src/IdentityInterface.php#L40 right?

LordSimal: I was just copying it from https://github.com/cakephp/authorization/blob/3.x/src/IdentityDecorator.php#L63

dereuromark: Yeah, but I guess new code could already use the full types.

dereuromark: Technically, even those classes are updatable, since mixed will be fully BC (max type) as well as not having it in extension seems also fine, as it currently already shows.

LordSimal: then let me adjust the PR

LordSimal: Just made sure this acutally works: https://onlinephp.io/c/5e865

dereuromark (approved):

PR #275, opened 2024-01-15
3.next: add DIC support to OrmResolver

LordSimal:
This is a proposal initiated by @jamisonbryant to make at least the OrmResolver more DI aware.

Instead of having to pass on services to `->can()` etc. one can easily inject the needed services into the policies which actually need them to do whatever logic is required to authorize the current user on the given action.

To get this working from a user perspective, all they would have to adjust in their `Application::getAuthorizationService` method is
```
$ormResolver = new OrmResolver();
```
to
```
$ormResolver = new OrmResolver('App', [], $this->getContainer());
```

After that all policies retrieved by the OrmResolver can leverage constructor injection.

If we want to go this way then we should also make the MapResolver DI aware as well.

### DIFF
--- a/src/Policy/OrmResolver.php
+++ b/src/Policy/OrmResolver.php
@@ -18,6 +18,7 @@ namespace Authorization\Policy;
 
 use Authorization\Policy\Exception\MissingPolicyException;
 use Cake\Core\App;
+use Cake\Core\ContainerInterface;
 use Cake\Datasource\EntityInterface;
 use Cake\Datasource\QueryInterface;
 use Cake\Datasource\RepositoryInterface;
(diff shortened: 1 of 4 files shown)


src/Policy/OrmResolver.php
```diff
@@ -146,7 +159,13 @@ protected function findPolicy(string $class, string $name, string $namespace): m
             throw new MissingPolicyException([$class]);
         }
 
-        return new $policyClass();
+        if ($this->container && $this->container->has($policyClass)) {
+            $policy = $this->container->get($policyClass);
```
jamisonbryant, line 163: Per the interface, `get()` can throw a variety of exceptions. Should this be handled by this method (i.e. fallback style) or leave it entirely to userland code?

LordSimal: Well `get($policyClass)` won't throw any more since we checked the container with `has($policyClass)`.

This is the same procedure as we do when [parsing controller action arguments
](https://github.com/cakephp/cakephp/blob/5.x/src/Controller/ControllerFactory.php#L166-L169) just with the difference that the container could not be set in our case to be as BC as possible.

jamisonbryant: Technically, `$this->container->get()` can still throw `ContainerExceptionInterface` however you're correct in that the `has()` check will prevent the `NotFoundException` from being thrown.

At least, that's my understanding from reading `ContainerInterface`. I checked out the ControllerFactory code and see what you mean...but I must ask: should we be catching/handling the container exception there as well?

LordSimal: We use the [league/container](https://github.com/thephpleague/container) as an implementation for the PSR-11 Container Interface. The only reason this exception could be thrown is [here](https://github.com/thephpleague/container/blob/4.x/src/Container.php#L195).

So technically, one could add a service provider which lies about providing a specific service.
But then I'd say its fine if we just let that exception bubble up to the top and not catch it because it's a configuration problem of the user (or a plugin), not something we should actively catch and handle for the user.

markstory:
Seems like a good change to me.

LordSimal:
@jamisonbryant anything to add from your side? If not I'd merge this and add support for DI to the MapResolver as well

jamisonbryant:
No, this looks good to me. Well done!

PR #277, opened 2024-01-21
3.x: stan update

LordSimal:


psalm-baseline.xml
```diff
@@ -1,2 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.x-dev@"/>
+<files psalm-version="5.20.0@3f284e96c9d9be6fe6b15c79416e1d1903dcfef4">
+  <file src="src/Command/PolicyCommand.php">
+    <RiskyTruthyFalsyComparison>
+      <code>empty($name)</code>
```
dereuromark: Couldnt this be reworked right away to not use cloaking maybe?

src/Command/PolicyCommand.php
```diff
@@ -76,7 +76,7 @@ public function templateData(Arguments $arguments): array
         $data = parent::templateData($arguments);
 
         $name = $arguments->getArgument('name');
-        if (empty($name)) {
+        if ($name === null || $name === '') {
```
LordSimal: I don't think this is better but psalm thinks it is 🤷🏻

dereuromark: `if (!$name)` was what I had in mind :)

LordSimal: but psalm is not happy about that one as I described in core chat.

Because then `null` and `''` would be handled in the same if which "could lead to unexpected behavior" [according to psalm](https://psalm.dev/docs/running_psalm/issues/RiskyTruthyFalsyComparison/).

dereuromark: Thus the silencing on top of it :)

dereuromark: Otherwise you would have to use 
```php
/** @var non-empty-string|null $name */
$name = ...
if ($name === null) 
```
if you wanted to make this clean

LordSimal: Well If I just look at https://www.php.net/manual/en/types.comparisons.php then we didn't really do anything here other than just suppress the warning...

Issue #265, opened 2023-11-10
Feature request: Make Authorization service available via DI in the Middleware


jamisonbryant:
Per [a recent community discussion](https://cakesf.slack.com/archives/C053DPNGT/p1699626768751309) it would be helpful for some apps to be able to obtain an instance of `AuthorizationServiceInterface` and/or the currently configured `AuthorizationService` from the service container.

Applications that rely heavily on the service container to build composable and reusable services often have to remain aware of Authorization concerns but do not want to rely on `ServerRequest` directly if avoidable. For example, a userland Searcher service class needs to scope a `Query` to members-only but cannot directly invoke the `AuthorizationService` without first obtaining it from a request attribute.

If the service were available via DI, that same Searcher class could list it as a dependency when configuring the application service container and then call it directly from within the service. This provides greater type safety and less (direct) reliance on the HTTP request object.

markstory:
I agree that having the AuthorizationService & AuthenticationServices added to the applications DI container by the plugins makes sense.

jamisonbryant:
One issue with the current implementation is that it assumes that the object providing the auth services is the same object providing the container, or at least that they have the same interfaces. This requires a bit of an awkward coupling/passing around of objects in our code as this is explicitly _not_ the case for us.

The docs say to attach the `AuthorizationServiceProviderInterface` to the `Application` class and then to register the middleware by calling `->add(new AuthorizationMiddleware($this))`. Well, we don't do that. Instead, we have a separate class entirely that implements the `AuthorizationServiceProviderInterface` and then we add the middleware by passing `new CustomAuthorizationServiceProvider()` instead of `$this`. We feel this is better separation of concerns.

But, since the `AuthorizationMiddleware` is checking for the `ContainerApplicationInterface` this means that we must implement this interface on our custom class in order to have the authz service available in DI. This further requires us to pass the container from the application through to the custom authz service provider just so the middleware can get it.

Here's what I mean:
```php
/* file: src/Application.php */

// in the middleware() method
->add(new AuthenticationMiddleware(
  // pass the DI container to the custom service provider so middleware can get to it
  new CustomAuthenticationServiceProvider($this->getContainer())
))

/* file: src/Service/Provider/CustomAuthorizationServiceProvider.php */

public function __construct(ContainerInterface $container)
{
  // this property and constructor exist solely so the AuthzMiddleware can access the DI container
  $this->appContainer = $container;
}

public function services(ContainerInterface $container): void
{
[...]

PR #278, opened 2024-01-22
2.next: more DIC

LordSimal:
yea... you read that right.... 😁 

This is a direct response to https://github.com/cakephp/authorization/issues/265#issuecomment-1877876524 and the need to set a custom DIC in the authorization handler if the passed down subject doesn't actually implement the `ContainerApplicationInterface`

This allows users to extract the container logic from the authorization service logic if they for some reason don't want it to have it all in their Application.php


jamisonbryant:
This looks good, and it enables us to remove a lot of awkward workaround code that we had to have before when we were forced to implement the `ContainerApplicationInterface`.

Thanks Kevin!

PR #283, opened 2024-03-08
Make AuthorizationService available in DIC

jamisonbryant:
See related PR #278 

I ran all the tools and checks except Psalm, it encountered a ton of deprecation errors from a dependency and crashed.

PR #284, opened 2024-03-13
Forward port DIC related changes done in 2.next to 3.next

ADmad:
Refs  #283


jamisonbryant (approved):
This looks good to me, and we have tested this locally. Thanks for the port.

PR #299, opened 2024-11-15
Update policies.rst

luizcmarin:
Bake generated "public function canEdit", not "public function canUpdate"


markstory:
Thank you 🎉

PR #300, opened 2024-11-17
update stan

LordSimal:


phpstan-baseline.neon
```diff
@@ -0,0 +1,37 @@
...
+			path: src/Controller/Component/AuthorizationComponent.php
+
+		-
+			message: '#^Method Authorization\\Controller\\Component\\AuthorizationComponent\:\:implementedEvents\(\) should return array\<string, mixed\> but returns array\<int\|string, string\>\.$#'
+			identifier: return.type
+			count: 1
+			path: src/Controller/Component/AuthorizationComponent.php
+
+		-
+			message: '#^Cannot call method getIdentifier\(\) on array\|Authentication\\IdentityInterface\.$#'
```
ADmad, line 22: This error should actually be addressed at some point, good for now.

phpstan-baseline.neon
```diff
@@ -0,0 +1,37 @@
...
+			path: src/Controller/Component/AuthorizationComponent.php
+
+		-
+			message: '#^Method Authorization\\Controller\\Component\\AuthorizationComponent\:\:canResult\(\) should return Authorization\\Policy\\ResultInterface but returns Authorization\\Policy\\ResultInterface\|bool\.$#'
+			identifier: return.type
+			count: 1
+			path: src/Controller/Component/AuthorizationComponent.php
+
+		-
+			message: '#^Method Authorization\\Controller\\Component\\AuthorizationComponent\:\:implementedEvents\(\) should return array\<string, mixed\> but returns array\<int\|string, string\>\.$#'
```
ADmad: Why does it think an array with `int` key is returned?

LordSimal: because `$this->getConfig('authorizationEvent')` returns mixed

phpstan-baseline.neon
LordSimal: we have that prop as `ArrayAccess|array` with native PHP attributes but overwritten with `@var \Authentication\IdentityInterface` because of the nature of it being a decorator implementation.

Haven't taken too much time thinking about it but feels pretty hard to get stan to understand that correctly.

phpstan-baseline.neon
dereuromark: String cast it

src/IdentityDecorator.php
```diff
@@ -111,6 +110,8 @@ public function __call(string $method, array $args): mixed
         }
         $call = [$this->identity, $method];
 
+        assert(is_callable($call), 'Method does not exist');
```
ADmad: I would prefer adding an inline ignore for the error instead of the `assert()`. PHP is gonna throw an error anyway if it's not a valid callable.

phpstan-baseline.neon
ADmad: Yes it's a tricky one, hence I too have been ignoring it :stuck_out_tongue:.

PR #303, opened 2025-01-23
fix SelectQuery namespace for Table policies

LordSimal:
As reported in https://discourse.cakephp.org/t/return-value-must-be-of-type-cake-orm-selectquery-on-tablepolicy/12396/1

`Cake\ORM\SelectQuery` does not exist, only `Cake\ORM\Query\SelectQuery`


markstory:
Oops. That's my mistake. Thank you for fixing it.

PR #306, opened 2025-06-09
Add test for mock model policy resolution and refactor `getRepository…

mirko-pagliai:
This was not possible:

```php
        $articles = $this->getMockForModel('Articles');
        $resolver = new OrmResolver('TestApp');
        $policy = $resolver->getPolicy($articles->find());
```
because it caused:
```
Authorization\Policy\Exception\MissingPolicyException: Policy for `MockObject_ArticlesTable_23778e41` has not been defined.
/home/mirko/Libs/authorization/src/Policy/OrmResolver.php:164
/home/mirko/Libs/authorization/src/Policy/OrmResolver.php:135
/home/mirko/Libs/authorization/src/Policy/OrmResolver.php:93
/home/mirko/Libs/authorization/tests/TestCase/Policy/OrmResolverTest.php:122
```

In other words, using a mock, with `get_class($table)` it was impossible to determine the corresponding policy.  
Using `$table->getAlias(). 'Table'` not only is the thing simplifies, but the mock of a model can be used.

I noticed that this can be useful for testing the Policy Scopes, not being able to use the actual model.

Example:
```php
class NotificationsTablePolicy
{
    public function scopeIndex(User $Identity, SelectQuery $Query): SelectQuery
    {
        return $Query->find('InnerJoinWithUserId', $User->id);
    }
}
```

A mock can now be used and expect that the `find()` method is called with those arguments.

### DIFF
--- a/src/Policy/OrmResolver.php
+++ b/src/Policy/OrmResolver.php
@@ -121,11 +121,10 @@ class OrmResolver implements ResolverInterface
      */
     protected function getRepositoryPolicy(RepositoryInterface $table): mixed
     {
-        $class = get_class($table);
         $tableNamespace = '\Model\Table\\';
+        $class = get_class($table);
+        $name = $table->getAlias() . 'Table';
         $namespace = str_replace('\\', '/', substr($class, 0, (int)strpos($class, $tableNamespace)));
-        /** @psalm-suppress PossiblyFalseOperand */
-        $name = substr($class, strpos($class, $tableNamespace) + strlen($tableNamespace));
 
         return $this->findPolicy($class, $name, $namespace);
     }
--- a/tests/TestCase/Policy/OrmResolverTest.php
+++ b/tests/TestCase/Policy/OrmResolverTest.php
@@ -115,6 +115,14 @@ class OrmResolverTest extends TestCase
         $this->assertInstanceOf(ArticlesTablePolicy::class, $policy);
     }
 
+    public function testGetPolicyMockModel()
+    {
+        $articles = $this->getMockForModel('Articles');
+        $resolver = new OrmResolver('TestApp');
+        $policy = $resolver->getPolicy($articles->find());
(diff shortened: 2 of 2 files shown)


mirko-pagliai:
Instead it does not work, because in any case that `get_class()` is still necessary, and the instance of a mock will be something like `Mockobject_Notificationstable_e144e823`.

Do something like:
```php
if ($table instanceof MockObject){
}
```
It would be a bad idea, because it would require phpunit as a dependence in production.

Instead, this works:
```php
if (!str_ends_with($class, 'Table') && str_ends_with(get_parent_class($table) ?: '', 'Table')) {
   $class = get_parent_class($table);
}
```

But obviously as an idea is definitely bad, I say it first.

Perhaps this implementation is not possible in a simple way.
For now I close, waiting to see if more comes to my mind.

mirko-pagliai:
In the end, the idea was good (allowing to solve the policy also starting from mocks). However, in hindsight the implementation, in addition to being complex, would be very "questionable".

After writing some code, I would say that the best solution is to intervene with a mock of the query (rather than the table).

So (I'm posting the solution, in case it's useful to someone else, even though the purpose of a repository is not to offer a solution, but as an alternative to PR)
```php
class NotificationsTablePolicy {
    public function scopeIndex(User $Identity, SelectQuery $Query): SelectQuery
    {
        return $Query->find('InnerJoinWithUserId', $Identity->id);
    }
}
```

```php
    public function testScopeIndex(): void
    {
        $Query = Mockery::mock(SelectQuery::class . '[find]', [$this->fetchTable('Notifications')]);
        $Query
            ->shouldReceive('find')
            ->once()
            ->with('InnerJoinWithUserId', 1)
            ->andReturnSelf();

        $this->User->applyScope('index', $Query);
    }
```

`$this->User` is obviously set elsewhere.  
Of course you could let a real query be created and make assertions on the generated sql, but this was the case where this is not convenient and it is better to verify the call to `find()`.

PR to be considered resolved as closed.

PR #308, opened 2025-06-27
Allow skipping redirects for extensions.

### LINKED ISSUE #717: RedirectHandler issues with non HTML
When using extension based routing and clearly non-HTML actions, if there are any permission issues
the user gets redirected to /login (302)

The URL e.g. is

    /updates/check.json

And the response is expected to be JSON.
So here the expectation would be for the handler to actually ignore the redirect part and just throw the 4xx exception I assume?
As there should be no redirects on such endpoints from the auth part.

dereuromark:
Resolves https://github.com/cakephp/authentication/issues/717

This happens when you have default (admin) routing and just add extensions into it in general.
You cannot then control for what actions only.
And if you then get redirects on those, that is weird (API etc) and creates overhead on the server.


src/Middleware/UnauthorizedHandler/RedirectHandler.php
```diff
@@ -44,6 +44,7 @@ class RedirectHandler implements HandlerInterface
         'url' => '/login',
         'queryParam' => 'redirect',
         'statusCode' => 302,
+        'allowedRedirectExtensions' => false,
```
markstory: We should add this new option to the docblock.

docs/en/middleware.rst
```diff
@@ -198,6 +198,10 @@ Both redirect handlers share the same configuration options:
 * ``queryParam`` - the accessed request URL will be attached to the redirect URL
   query parameter (``redirect`` by default).
 * ``statusCode`` - HTTP status code of a redirect, ``302`` by default.
+* ``allowedRedirectExtensions`` - an array of allowed file extensions for redirecting.
+  If the request URL has a file extension that is not in this list, the redirect will not
+  happen and the exception will be rethrown. This is useful to prevent unauthorized access
+  to API based responses, that should not be redirecting in any case. `false` by default and not enabled then.
```
markstory:
```suggestion
* ``allowedRedirectExtensions`` - an array of allowed file extensions for redirecting.
  If the request URL has a file extension that is not in this list, the redirect will not
  happen and the exception will be rethrown. Can also be a boolean to toggle on/off
  redirects entirely. This is useful to prevent unauthorized access to API based 
  responses, that should not be redirecting in any case. `false` by default and not enabled then.
```

tests/TestCase/Middleware/UnauthorizedHandler/RedirectHandlerTest.php
```diff
@@ -160,4 +160,48 @@ public function testHandleException()
...
+        $handler->handle($exception, $request, [
+            'exceptions' => [
+                Exception::class,
+            ],
+            'url' => '/users/login',
+            'allowedRedirectExtensions' => [],
+        ]);
+    }
+
+    public function testHandleRedirectionWithExtensionWhitelisted(): void
```
markstory:
```suggestion
    public function testHandleRedirectionWithExtensionAllowlisted(): void
```

tests/TestCase/Middleware/UnauthorizedHandler/RedirectHandlerTest.php
```diff
@@ -160,4 +160,48 @@ public function testHandleException()
...
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_METHOD' => 'GET'],
+        );
+        $request = $request->withParam('_ext', 'csv');
+
+        $this->expectException(Exception::class);
+
+        $handler->handle($exception, $request, [
+            'exceptions' => [
+                Exception::class,
```
markstory:
```suggestion
        $exception = new LogicException();
        $request = ServerRequestFactory::fromGlobals(
            ['REQUEST_METHOD' => 'GET'],
        );
        $request = $request->withParam('_ext', 'csv');

        $this->expectException(LogicException::class);

        $handler->handle($exception, $request, [
            'exceptions' => [
                LogicException::class,
```

Any concern on this being a broad type expectation? You could use something more narrow `LogicException` and have more tightly scoped behavior captured in the test.

src/Middleware/UnauthorizedHandler/RedirectHandler.php
dereuromark: Done
I switched the default to true, as that is more logical.

src/Exception/LogicException.php
```diff
@@ -0,0 +1,70 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authorization\Exception;
+
+use Authorization\Policy\ResultInterface;
+use Throwable;
+
+class LogicException extends Exception
```
markstory: I thought you would have used the stdlib `LogicException` but this works too. :shrug:  Edit: I see now that my suggestion is not possible. I'll push a commit up instead of leaving more comments.

src/Exception/LogicException.php
```diff
@@ -0,0 +1,70 @@
...
+
+        parent::__construct($message, $code, $previous);
+    }
+
+    /**
+     * Returns policy check result if passed to the exception.
+     *
+     * @return \Authorization\Policy\ResultInterface|null
+     */
+    public function getResult(): ?ResultInterface
```
markstory: I don't think this method will be used.

markstory (approved):

PR #322, opened 2026-03-17
Allow Closure for requireAuthorizationCheck config option

dereuromark:
## Summary

This PR adds support for a callable in the `requireAuthorizationCheck` middleware configuration option. Currently, this option only accepts a boolean value, but applications often need to conditionally skip authorization checks based on the request path or other request attributes.

## Use Case

When integrating third-party plugins that provide their own admin panels (e.g., queue management dashboards), these plugins may manage authorization independently. The host application needs a way to skip the middleware's authorization check for these specific routes without disabling it globally.

**Before this PR**, applications had to wrap the middleware in a custom closure:

```php
$middlewareQueue->add(function ($request, $handler) use ($app) {
    $path = $request->getUri()->getPath();
    $skipAuthCheck = str_contains($path, '/admin/queue');

    $middleware = new AuthorizationMiddleware($app, [
        'requireAuthorizationCheck' => !$skipAuthCheck,
    ]);

    return $middleware->process($request, $handler);
});
```

**After this PR**, the same can be achieved cleanly:

```php
$middlewareQueue->add(new AuthorizationMiddleware($this, [
    'requireAuthorizationCheck' => function ($request) {
        $path = $request->getUri()->getPath();
        if (str_contains($path, '/admin/queue')) {
            return false;
        }
        return true;
    }
]));
```

## Changes

* Modified `AuthorizationMiddleware::process()` to check if `requireAuthorizationCheck` is callable and invoke it with the request
* Updated docblock to document the callable signature
* Added 3 test cases covering:
  - Callable returning `true` (requires authorization check)
  - Callable returning `false` (skips authorization check)
  - Route-based logic example
* Updated English documentation with example usage

## Backwards Compatibility

This change is fully backwards compatible. The existing boolean behavior is preserved - the callable is only invoked if the config value is callable.

### DIFF
--- a/src/Middleware/AuthorizationMiddleware.php
+++ b/src/Middleware/AuthorizationMiddleware.php
@@ -30,6 +30,7 @@ use Authorization\Middleware\UnauthorizedHandler\UnauthorizedHandlerTrait;
 use Cake\Core\ContainerApplicationInterface;
 use Cake\Core\ContainerInterface;
 use Cake\Core\InstanceConfigTrait;
+use Closure;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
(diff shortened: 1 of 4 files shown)


src/Middleware/AuthorizationMiddleware.php
```diff
@@ -135,7 +136,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
         try {
             $response = $handler->handle($request);
 
-            if ($this->getConfig('requireAuthorizationCheck') && !$service->authorizationChecked()) {
+            $requireCheck = $this->getConfig('requireAuthorizationCheck');
+            if (is_callable($requireCheck)) {
```
dereuromark: @ADmad Should this be Closure as well?

LordSimal:
As this is a new feature shouldn't this be added to 3.next?

dereuromark:
For these small satellite repos its sometimes overkill, since this would be a simple minor release without all the branch switching/merging.
But I did it now cleanly.

ADmad (approved):

PR #331, opened 2026-05-12
Support sub-directory entities and tables in OrmResolver

### LINKED ISSUE #315: Issue with sub-directories
I recently installed this package to slowly enable authorization for my project. Installing it and configuring the basics worked perfectly fine. However, I'm a bit struggling with implementing the policy/entity relation using the `OrmResolver`.
 
I'm not sure if I'm doing something wrong here, but it seems that placing entities/policies in sub-directories doesn't seem to work for me.

My folder structure:
```
src/
    Model/
        Entity/
            SubDir/
                User.php
    Policy/
        SubDir/
            UserPolicy.php
```

I'm using the `OrmResolver` to automatically resolve the entity/policy relation as my project has a lot of entities (~170). Using the `MapResolver` to map the relations manually would be quite an inconvenience.

I noticed during debugging that my issue mainly lies within the `OrmResolver::getEntityPolicy` or `App::className`.
The first function seems to prepare the data for `App::className` to resolve the Policy based on an entity instance/class.

It resolves to something like this:
```
App::className('App.SubDir\User', 'Policy', 'Policy');
// = null
```

The implementation of `App::className` has an early return on the first line (`str_contains($class, '\\')`) which stops the search directly.

If the usage would be changed to the following, the `App::className` would be working as intended.
```
App::className('App.SubDir/User', 'Policy', 'Policy'); // Notice the '\' being changed to '/'.
// = 'App\Policy\SubDir\UserPolicy'
```

I can't tell if this is an issue within the core cakephp function, or due to the preparation done by `OrmResolver::getEntityPolicy`.
Does anyone know if this is a simple misconfiguration on my end, a bug in this project or a bug in the cakephp project?

dereuromark:
Closes #315.

## Background

The `OrmResolver` already converts the **namespace** segment of a resource class from backslashes to forward slashes before handing it to `App::className()`:

```php
$namespace = str_replace('\\', '/', substr($class, 0, (int)strpos($class, $entityNamespace)));
```

…but the **name** segment is left as-is. For a class like `TestApp\Model\Entity\SubDir\Widget` the resolver therefore passes `App.SubDir\Widget` to `App::className()`. `App::className()` short-circuits on any embedded backslash and returns `null`, so a `MissingPolicyException` is thrown even when `App\Policy\SubDir\WidgetPolicy` exists at the expected location.

The namespace side of this conversion was added intentionally — so partial sub-directory support is already present. This change finishes the job for the name half.

## Change

Apply the same `str_replace('\\', '/', …)` to the name segment in both `getEntityPolicy()` and `getRepositoryPolicy()`.

## Note on previous discussion

Issue #315 was closed as "use FQCN" — that workaround still works (via `MapResolver` registration), and consumers who want manual control can keep doing it. This PR is additive: nothing changes for flat layouts, and the same convention `App\Policy\SubDir\WidgetPolicy` for `App\Model\Entity\SubDir\Widget` simply starts working. If maintainers prefer to keep the convention-based resolver flat-only, feel free to close.

## Verification

- `composer test` — 142 tests, 291 assertions (was 140 / 289). 2 new tests, one entity and one table, both with a `SubDir/Widget(s)` fixture.
- `composer stan` — clean.
- `composer cs-check` — clean.

### DIFF
--- a/src/Policy/OrmResolver.php
+++ b/src/Policy/OrmResolver.php
@@ -103,7 +103,7 @@ class OrmResolver implements ResolverInterface
         $class = $entity::class;
         $entityNamespace = '\Model\Entity\\';
         $namespace = str_replace('\\', '/', substr($class, 0, (int)strpos($class, $entityNamespace)));
-        $name = substr($class, strpos($class, $entityNamespace) + strlen($entityNamespace));
+        $name = str_replace('\\', '/', substr($class, strpos($class, $entityNamespace) + strlen($entityNamespace)));
 
         return $this->findPolicy($class, $name, $namespace);
     }
@@ -119,7 +119,7 @@ class OrmResolver implements ResolverInterface
         $class = $table::class;
         $tableNamespace = '\Model\Table\\';
         $namespace = str_replace('\\', '/', substr($class, 0, (int)strpos($class, $tableNamespace)));
-        $name = substr($class, strpos($class, $tableNamespace) + strlen($tableNamespace));
+        $name = str_replace('\\', '/', substr($class, strpos($class, $tableNamespace) + strlen($tableNamespace)));
 
         return $this->findPolicy($class, $name, $namespace);
     }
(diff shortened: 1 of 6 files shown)


markstory:
> If maintainers prefer to keep the convention-based resolver flat-only, feel free to close.

Aren't you also a maintainer? :confused:

markstory (approved):
Seems fine to me.

